PR #3, opened 2015-09-23
Add Travis config

ceedubs:
And update to Scala 2.11.7

After merging this I think you should just have to go to
https://travis-ci.org/profile/stew and flip the switch for dogs.


stew:
sweet, TYVM

PR #5, opened 2015-09-24
Immutable, singly-linked list implementation.

non:
This code is very similar to scala.List, with a few key differences:

1. It does not expose any "unsafe" methods.
2. It is invariant, whereas scala.List is covariant.
3. It uses subtyping to differentiate non-emptiness.
4. It provides Cats type class instances.
5. It does not currently provide a fast (mutable) builder.

The types defined here are as follows:

 - Lst[A] represents a list of zero-or-more elements.
 - Nel[A] represents a list of one-or-more elements.
 - El[A]  represents an empty list (exactly zero elements).

(Every Lst[A] is either a Nel[A] or an El[A].)

While it does not (yet) provide every single Scala collection
method, it provides a decent number of them. Type class instances
are provided for both Lst and Nel (for example, Nel has a Comonad
instance, whereas Lst only has a CoflatMap instance).


non:
For what it's worth, I think the names `Lst`, `Nel`, and `El` are probably not great. I couldn't think of anything compelling enough to use instead. I think before this is merged we should figure out what to call these things.

I think it's relatively important that at least `Lst` and `Nel` both look like reasonable type names that users won't mind typing int their code.

stew:
I'm going to write up some tests for this, but I like the idea, It is one I have played around with in the past.

These names aren't terrible. "a list is either a nel or an el" has a nice poetic meter :). I'd also be happy with Cons and End instead of Nel and El

I am able to convince myself that I would be okay with calling the trait `List`. Which is certainly a bad idea because its going to piss a bunch of people off right out of the gate, so perhaps, nevermind.

src/main/scala/Lst.scala
```diff
@@ -0,0 +1,282 @@
...
+      case _ => false
+    }
+
+  final def cata[B](b: => B, f: (A, Lst[A]) => B): B =
+    this match {
+      case Nel(h, t) => f(h, t)
+      case El() => b
+    }
+
+  final def getNonEmpty: Option[Nel[A]] =
```
julien-truffaut: `toNel` would seem more natural to me

src/main/scala/Lst.scala
```diff
@@ -0,0 +1,282 @@
...
+
+  @tailrec final def foldLeft[B](b: B)(f: (B, A) => B): B =
+    this match {
+      case El() => b
+      case Nel(h, t) => t.foldLeft(f(b, h))(f)
+    }
+
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B]
+
+  def :::(as: Lst[A]): Lst[A]
```
julien-truffaut: why some method pattern match on `this` and other delegate to child?

non: In general, I prefer using pattern matching, but I used delegation anywhere that the return type for the `Nel` and `Lst` versions are different.

If you'd prefer consistency, I could make all of the methods use delegation.

src/main/scala/Lst.scala
non: That's a good idea, I'll change it.

stew:
I'm super busy this week, and there is no reason for this thing to sit out on a PR, I'm going to just go ahead and merge for now, 

When I'm above water again, I'll try to convert the things in Dequeue to Lst to see how natural that feels and add the rest of the law checking.

PR #9, opened 2016-01-29
Feature/nocats

anicolaspp:
I added a Binary Tree to start with something simple. It does contain its test as its being written using TDD.

### DIFF
--- a/core/src/main/scala/BinaryTree.scala
+++ b/core/src/main/scala/BinaryTree.scala
@@ -4,11 +4,14 @@
 
 package dogs
 
+import javafx.scene.transform.Rotate
+
 import Predef._
 
 import dogs.Order.{GT, EQ, LT}
 
 import scala.annotation.tailrec
+import scala.math
 import scala.List._
 
 
@@ -19,64 +22,116 @@ abstract class BinaryTree[A] {
   val left: BinaryTree[A]
   val right: BinaryTree[A]
 
+  def balanceFactor: Int
+
+  private[dogs] def update(): BinaryTree[A]
 
   def isEmpty: Boolean
 
   def toLst(): List[A] = this match {
     case BTNil()                =>  El[A]
-    case Branch(Some(a), l, r)  =>  l.toLst ::: (a :: r.toLst)
+    case Branch(Some(a), l, r, _, _)  =>  l.toLst ::: (a :: r.toLst)
   }
 
   def inorder() = toLst()
 
   def preorder(): List[A] = this match {
     case BTNil()                =>  El[A]
-    case Branch(Some(a), l, r)  =>  a :: l.preorder ::: r.preorder
+    case Branch(Some(a), l, r, _, _)  =>  a :: l.preorder ::: r.preorder
   }
 
   def posorder(): List[A] = this match {
     case BTNil()                =>  El[A]
-    case Branch(Some(a), l, r)  =>  l.posorder() ::: r.posorder() ::: Nel(a, El[A])
+    case Branch(Some(a), l, r, _, _)  =>  l.posorder() ::: r.posorder() ::: Nel(a, El[A])
   }
 
   def min(): Option[A] =  {
     @tailrec def loop(sub: BinaryTree[A], x: Option[A]): Option[A] = sub match {
       case BTNil()              =>  x
-      case Branch(a, l, _)      =>  loop(l, a)
+      case Branch(a, l, _, _, _)      =>  loop(l, a)
     }
 
     this match {
       case BTNil()              => None()
-      case Branch(a, l, _)      =>  loop(l, a)
+      case Branch(a, l, _, _, _)      =>  loop(l, a)
     }
   }
 
   def max(): Option[A] = {
     @tailrec def loop(sub: BinaryTree[A], x: Option[A]): Option[A] = sub match {
       case BTNil()              =>  x
-      case Branch(a, _, r)      =>  loop(r, a)
+      case Branch(a, _, r, _, _)      =>  loop(r, a)
     }
 
     this match {
       case BTNil()              =>  None()
-      case Branch(a, _, r)      =>  loop(r, a)
+      case Branch(a, _, r, _, _)      =>  loop(r, a)
     }
   }
 
-  def add(x: A)(implicit order: Order[A]): Branch[A] = this match {
-    case BTNil()                =>  Branch(Some(x), BinaryTree.empty, BinaryTree.empty)
-    case Branch(Some(a), l, r)  =>  order.compare(x, a) match {
-      case LT                     =>  Branch(Some(a), l.add(x), r)
-      case EQ                     =>  this.asInstanceOf[Branch[A]]
-      case GT                     =>  Branch(Some(a), l, r.add(x))
+  def balance(tree: BinaryTree[A]): BinaryTree[A] = {
+    var t = tree.update
+
+    if (t.balanceFactor > 1) {
+      if (t.right.balanceFactor <= 0) {
+        t = Branch(t.value, t.left, rotate(t.right, 1), t.size, t.height)
+      }
+      t = rotate(t, -1)
+    } else {
+      if (t.balanceFactor < -1)
+      {
+        if (t.left.balanceFactor >= 0){
+          t = Branch(t.value, rotate(t.left, -1), t.right, t.size, t.height)
+        }
+        t = rotate(t, 1)
+      }
+    }
+
+    return t
+  }
+
+  def rotate(x: BinaryTree[A], direction: Int): BinaryTree[A] = {
+    if (direction < 0) {  // left
+
+      if (x == BTNil() || x.right == BTNil()) return x
+
+      val y = x.right
+      val a = Branch(x.value, x.left, y.left, x.size, x.height).update()
+
+      val b = Branch(y.value, a, y.right, y.size, y.height).update()
+
+      return b
+    }
+    else { //right
+      if (x == BTNil() || x.left == BTNil()) return x
+
+      val y = x.left
+      val a = Branch(x.value, BTNil(), y.right, x.size, x.height).update()
+
+      val b = Branch(y.value, y.left, a).update()
+
+      return b
     }
   }
 
-  def +(x: A)(implicit order: Order[A]): Branch[A] = add(x)
+  def add(x: A)(implicit order: Order[A]): Branch[A] = {
+    def insert(x: A, order: Order[A])  = this match {
+      case BTNil()                      =>  Branch(Some(x), BinaryTree.empty, BinaryTree.empty)
+      case Branch(Some(a), l, r, _, _)  =>  order.compare(x, a) match {
+        case LT                     =>  Branch(Some(a), l.add(x)(order), r)
+        case EQ                     =>  this.asInstanceOf[Branch[A]]
+        case GT                     =>  Branch(Some(a), l, r.add(x)(order))
+      }
+    }
+
+    balance(insert(x, order)).asInstanceOf[Branch[A]]
+  }
+
+  def +(x: A)(implicit order: Order[A]): BinaryTree[A] = add(x)
 
   def remove(x: A)(implicit order: Order[A]): BinaryTree[A] = this match {
     case BTNil()                =>  BinaryTree.empty
-    case Branch(Some(a), l, r)  =>  order.compare(x, a) match {
+    case Branch(Some(a), l, r, _, _)  =>  order.compare(x, a) match {
       case LT =>  Branch(Some(a), l.remove(x), r)
       case GT =>  Branch(Some(a), l, r.remove(x))
       case EQ => (l, r) match {
@@ -95,7 +150,6 @@ abstract class BinaryTree[A] {
     }
   }
 
-
   def join(another: BinaryTree[A])(implicit order: Order[A]) = {
 
     @tailrec def build(sub: BinaryTree[A], xs: List[A]): BinaryTree[A] = xs match {
@@ -110,18 +164,50 @@ abstract class BinaryTree[A] {
   }
 
   def ++(another: BinaryTree[A])(implicit order: Order[A]) = join(another)
+
+
+  val size: Int
+
+  val height: Int
+
 }
 
-object BinaryTree {
-  def apply[A](a: A): BinaryTree[A] = Branch(Some(a),empty,empty)
 
-  def empty[A]: BinaryTree[A] = BTNil()
-}
 
 case class Branch[A](value: Option[A],
-                      left: BinaryTree[A],
-                      right: BinaryTree[A]) extends BinaryTree[A] {
+                     left: BinaryTree[A],
+                     right: BinaryTree[A],
+                     val size: Int = 0,
+                     val height: Int = 0) extends BinaryTree[A] {
 
+  override def balanceFactor: Int = this match {
+    case Branch(_, BTNil(), BTNil(), _, _)    =>  0
+    case Branch(_, BTNil(), r, _, _)          =>  r.asInstanceOf[Branch[A]].height
+    case Branch(_, l, BTNil(), _, _)          =>  -1 * l.asInstanceOf[Branch[A]].height
+    case Branch(_, l, r, _, _)                =>  r.asInstanceOf[Branch[A]].height - l.asInstanceOf[Branch[A]].height
+  }
+
+  override private [dogs] def update(): BinaryTree[A] = this match {
+    case Branch(a, BTNil(), BTNil(), _, _)  =>  Branch(a, BTNil(), BTNil(), 1, 1)
+    case Branch(a, BTNil(), r, _, _)        =>  {
+      val s = 1 + r.asInstanceOf[Branch[A]].size
+      val h = 1 + r.asInstanceOf[Branch[A]].height
+
+      Branch(a, BTNil(), r, s, h)
+    }
+    case Branch(a, l, BTNil(), _, _)        =>  {
+      val s = 1 + l.asInstanceOf[Branch[A]].size
+      val h = 1 + l.asInstanceOf[Branch[A]].height
+
+      Branch(a, l, BTNil(), s, h)
+    }
+    case Branch(a, l, r, _, _)              =>  {
+      val s = l.asInstanceOf[Branch[A]].size + r.asInstanceOf[Branch[A]].size
+      val h = 1 + math.max(l.asInstanceOf[Branch[A]].height, r.asInstanceOf[Branch[A]].height)
+
+      Branch(a, l, r, s, h)
+    }
+  }
 
 
   override def isEmpty: Boolean = false
@@ -129,6 +215,8 @@ case class Branch[A](value: Option[A],
   def apply(): BinaryTree[A] = this.asInstanceOf[BinaryTree[A]]
 
   def apply[B](a: B) = Branch[B](Some(a), BTNil(), BTNil())
+
+
 }
 
 case object BTNil extends BinaryTree[Nothing] {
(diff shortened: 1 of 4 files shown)


stew:
cool :)

code looks good. However, without any rebalacing, searching / inserting / etc is linear time (O(n)) not the O(log n) time you'd hope for with a balanced binary tree, right?

Is the plan to later add rebalancing to make this an AVL tree?

anicolaspp:
It is not balanced, I was thinking in create an AVL as a separated structure. If you don't think its a good idea to have both, I can balance that one. What do you think is best, keep this one and add AVL or just have one?

anicolaspp:
The other option is to apply DSW algorithm to convert this one into a AVL when need it. It is efficient and at the same time we can have both. I find not all the time we need an AVL, that why I prefer both. What do you think?

anicolaspp:
I fact, I will be keeping the Binary Tree and implementing the AVL as a subtype of it

stew:
Honestly I think we should only have  a balanced on if we were going to have one. If you have the choice of an unbalanced and a balanced tree, would you ever pick the unbalanced one?

anicolaspp:
I have seen some cases where the balanced one is not required, but its fine, I will be extending the unbalanced one and then we can only expose the balanced and keep the unbalanced private to the library. I hope that I have enough time this weekend to finish it up.

anicolaspp:
The Binary Tree is already balancing itself when add / remove.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,251 @@
...
+                     right: BinaryTree[A],
+                     private [dogs] val size: Int = 0,
+                     private[dogs] val height: Int = 0) extends BinaryTree[A] {
+
+  private [dogs] override def balanceFactor: Int = this match {
+    case Branch(_, BTNil(), BTNil(), _, _)    =>  0
+    case Branch(_, BTNil(), r, _, _)          =>  r.asInstanceOf[Branch[A]].height
+    case Branch(_, l, BTNil(), _, _)          =>  -1 * l.asInstanceOf[Branch[A]].height
+    case Branch(_, l, r, _, _)                =>  r.asInstanceOf[Branch[A]].height - l.asInstanceOf[Branch[A]].height
+  }
```
stew: couldn't we move height to be a method on BinaryTree, make the height of BTNil be 0, then we don't have to bother with asInstanceOf?

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,251 @@
...
+
+}
+
+
+
+case class Branch[A](value: Option[A],
+                     left: BinaryTree[A],
+                     right: BinaryTree[A],
+                     private [dogs] val size: Int = 0,
```
stew: shoudl height/size really be memoized? or should they just be calculated when needed?

in either case, shouldn't we take it out of the case class members, we never seem to use it in the patterm match

core/src/main/scala/BinaryTree.scala
anicolaspp: This can be done, I will take care of it. The problem was initially they were not in the abstract tree, but I did extract them but not change the section since all tests passed.

core/src/main/scala/BinaryTree.scala
anicolaspp: I cannot remove this field from here since sometimes they need to be initialized with a special values. Anyways they are private to the package.

We can recalculate them all the time, but during balancing they will be needed it a lot of time since balancing is recursive the same value will be calculated a lot of time. I rather have them calculated when going up in the tree and have them ready for next time

stew: The Branch constructor should be private right? We don't want people directly creating a branch, right?

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,251 @@
...
+      case GT                             =>  r.contains(x)
+    }
+  }
+
+  def add(x: A)(implicit order: Order[A]): Branch[A] = {
+    def insert(x: A, order: Order[A])  = this match {
+      case BTNil()                    =>  Branch(Some(x), BinaryTree.empty, BinaryTree.empty)
+      case Branch(Some(a), l, r, _, _)=>  order.compare(x, a) match {
+        case LT                     =>  Branch(Some(a), l.add(x)(order), r)
+        case EQ                     =>  this.asInstanceOf[Branch[A]]
```
stew: can you tell I really am uncomfortable with asInstanceeOf yet?  :)

you can change

    case Branch(Some(a)...

to 
    
    case b @ Branch(Some(a)....

then instead of this.asInstanceOf, you can return b

core/src/main/scala/BinaryTree.scala
anicolaspp: True, I will take care of it

core/src/main/scala/BinaryTree.scala
anicolaspp: Got it. There are few Scala stuff O dont know yet

anicolaspp: Others I just dont care about since my tests pass, that is my source of true.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,248 @@
...
+
+    another match {
+      case BTNil()              =>  this
+      case _                    =>  build(this, another.toLst())
+    }
+  }
+
+  def ++(another: BinaryTree[A])(implicit order: Order[A]) = join(another)
+
+  private def balance(tree: BinaryTree[A]): BinaryTree[A] = {
```
stew: balance is actaully undefined for BTNil, right? it really should be a Branch => Branch, right? what actually happens if you pass a BTNil in here?

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,248 @@
...
+    var t = tree.update
+
+    if (t.balanceFactor > 1) {
+      if (t.right.balanceFactor <= 0) {
+        t = Branch(t.value, t.left, rotate(t.right, 1), t.size, t.height)
```
stew: this statement has no effect, right? you are just going to overwrite the value of t with the very next statement, right?

stew: uh, no, i see. this is why I hate side-effects

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,248 @@
...
+      }
+    }
+
+    return t
+  }
+
+  private def rotate(x: BinaryTree[A], direction: Int): BinaryTree[A] = {
+    if (direction < 0) {  // left
+
+      if (x == BTNil() || x.right == BTNil()) return x
```
stew: no no no, never a return statement, never an if without an else.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,248 @@
...
+                     private[dogs] val height: Int = 0) extends BinaryTree[A] {
+
+  private [dogs] override def balanceFactor: Int = this match {
+    case Branch(_, BTNil(), BTNil(), _, _)    =>  0
+    case Branch(_, BTNil(), r, _, _)          =>  r.height
+    case Branch(_, l, BTNil(), _, _)          =>  -1 * l.height
+    case Branch(_, l, r, _, _)                =>  r.height - l.height
+  }
+
+  private [dogs] override def update(): BinaryTree[A] = this match {
```
stew: this method is undefined for BTNil and it only returns Branch, it never returns BinaryTree, right?

core/src/main/scala/BinaryTree.scala
anicolaspp: balance in fact, could receive a BTNIL. When remove if we trying to remove from the empty tree, BTNIL is passed to balance (with no effect of course) but I don't want to do the distinction on the remove method.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+    if (direction < 0) {  // left
+
+      if (x == BTNil() || x.right == BTNil()) {
+        x
+      }
+      else {
+        val y = x.right
+        val a = Branch(x.value, x.left, BTNil(), x.size, x.height).update()
+
+        Branch(y.value, a, y.right, y.size, y.height).update()
```
stew: y.left is thrown away here, right?

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+    }
+    else { //right
+      if (x == BTNil() || x.left == BTNil()) {
+        x
+      }
+      else {
+        val y = x.left
+        val a = Branch(x.value, BTNil(), x.right, x.size, x.height).update()
+
+        Branch(y.value, y.left, a, y.size, y.height).update()
```
stew: y.right is thrown away here

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+          t = Branch(t.value, rotate(t.left, -1), t.right, t.size, t.height)
+        }
+        t = rotate(t, 1)
+      }
+    }
+
+    t
+  }
+
+  private def rotate(x: BinaryTree[A], direction: Int): BinaryTree[A] = {
```
stew: why int for direction? are there more than 2 directions? this allows for 2^32 possible directions

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+                     right: BinaryTree[A],
+                     private [dogs] val size: Int = 0,
+                     private[dogs] val height: Int = 0) extends BinaryTree[A] {
+
+  private [dogs] override def balanceFactor: Int = this match {
+    case Branch(_, BTNil(), BTNil(), _, _)    =>  0
+    case Branch(_, BTNil(), r, _, _)          =>  r.height
+    case Branch(_, l, BTNil(), _, _)          =>  -1 * l.height
+    case Branch(_, l, r, _, _)                =>  r.height - l.height
+  }
```
stew: this is all just r.height - l.height right? I think all 4 case just end up being this.

tests/src/test/scala/BinaryTreeTest.scala
```diff
@@ -0,0 +1,235 @@
...
+
+class BinaryTreeTestAdd extends FlatSpec with Matchers {
+
+   implicit val cmp =  new Order[Int] {
+    override def apply(l: Int, r: Int): Ordering = l.compareTo(r) match {
+      case 0  => Order.EQ
+      case -1 => Order.LT
+      case 1  => Order.GT
+    }
+  }
```
stew: btw, dogs.std.intOrder exists

tests/src/test/scala/BinaryTreeTest.scala
```diff
@@ -0,0 +1,235 @@
...
+
+}
+
+class BinaryTreeTestJoins extends FlatSpec with Matchers {
+  implicit val cmp =  new Order[Int] {
+    override def apply(l: Int, r: Int): Ordering = l.compareTo(r) match {
+      case 0  => Order.EQ
+      case -1 => Order.LT
+      case 1  => Order.GT
+    }
```
stew: this looks familiar

tests/src/test/scala/BinaryTreeTest.scala
```diff
@@ -0,0 +1,235 @@
...
+  }
+}
+
+class BinaryTreeTestWalk extends FlatSpec with Matchers {
+  implicit val cmp =  new Order[Int] {
+    override def apply(l: Int, r: Int): Ordering = l.compareTo(r) match {
+      case 0  => Order.EQ
+      case -1 => Order.LT
+      case 1  => Order.GT
+    }
```
stew: this looks familiar

tests/src/test/scala/BinaryTreeTest.scala
```diff
@@ -0,0 +1,235 @@
...
+    sorted should be (tree.toLst)
+  }
+
+  it  should "return node-left-right for each node when walk in pre-order" in {
+    val tree = BinaryTree(5).add(2).add(1).add(3)
+
+    val preorder = tree.preorder
+
+    ScalaListConverter.toScalaList(preorder) should contain inOrder(2, 1, 5, 3)
+  }
```
stew: this seems very arbitrary, it seems like you crafted the test to make sure it passes, there are plenty of other ordered balanced trees that wouldn't have this ordering.

You should be crafting tests which make sure that don't know about what the implementation of the algorithm is, but instead just test that any invariants the algorithm have still hold.

so in this case we are testing that inserting elements in a certain order got some exact tree. but why do we care? all we should care about is that the tree is ordered and balanced.

tests/src/test/scala/BinaryTreeTest.scala
```diff
@@ -0,0 +1,235 @@
...
+
+  it should "insert value into left branch when add to brach to the left" in {
+
+    val branch = BinaryTree(5)
+
+    val result = branch.add(3)
+
+    result.value should be(Some(5))
+    result.left.value should be(Some(3))
+  }
```
stew: I'm not sure why we care if it is Branch(3, _, Branch(5, _, _) vs. Branch(5, Branch(3, _, _), _)

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+import scala.math
+
+
+
+abstract class BinaryTree[A] {
+
+  def value: Option[A]
+
+  val left: BinaryTree[A]
+  val right: BinaryTree[A]
```
stew: what is the usefulness of these to someone? are these two methods just useful in order to write tests?

seems like we don't want someone actually trying to traverse the tree themselves, we just want to give them combinators that work according to our invariants, that is that they will work in order, and that they will have a predictable computational complexity

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+  def preorder(): List[A] = this match {
+    case BTNil()                      =>  El[A]
+    case Branch(Some(a), l, r, _, _)  =>  a :: l.preorder ::: r.preorder
+  }
+
+  def posorder(): List[A] = this match {
+    case BTNil()                      =>  El[A]
+    case Branch(Some(a), l, r, _, _)  =>  l.posorder() ::: r.posorder() ::: Nel(a, El[A])
+  }
+
```
stew: are preorder and postorder actually useful? what is a use case?

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+    def insert(x: A, order: Order[A])  = this match {
+      case BTNil()                              =>  Branch(Some(x), BinaryTree.empty, BinaryTree.empty)
+      case branch @ Branch(Some(a), l, r, _, _) =>  order.compare(x, a) match {
+        case LT                     =>  Branch(Some(a), l.add(x)(order), r)
+        case EQ                     =>  branch
+        case GT                     =>  Branch(Some(a), l, r.add(x)(order))
+      }
+    }
+
+    balance(insert(x, order)).asInstanceOf[Branch[A]]
```
stew: seem like it would be much better to only ever try to balance branches and not trees which might be empty. I'd rather have the extra match on the delete case than have the .asInstanceOf here.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  def +(x: A)(implicit order: Order[A]): BinaryTree[A] = add(x)
+
+  def remove(x: A)(implicit order: Order[A]): BinaryTree[A] = {
+    def del(x:A)(implicit order: Order[A]): BinaryTree[A] = this match {
```
stew: no reason to pass either argument to this function, they are the same as the outer function always. the x in remove(x: A) is always the same as the x in del(x: A), so there is no reason to make this an argument to del. the same goes for order.

And after eliminating those two, there is no reason to have a nested def (since it needs no arguments).

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+
+      Branch(a, l, r, s, h)
+    }
+  }
+
+
+  override def isEmpty: Boolean = false
+
+
+  def apply[B](a: B) = Branch[B](Some(a), BTNil(), BTNil())
```
stew: I don't understand why this is here, It doesn't make sense to me.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+  private [dogs] val size: Int
+  private[dogs] val height: Int
+  private [dogs] def balanceFactor: Int
+  private[dogs] def update(): BinaryTree[A]
+
+  def isEmpty: Boolean
+
+  def toLst(): List[A] = this match {
+    case BTNil()                      =>  El[A]
+    case Branch(Some(a), l, r, _, _)  =>  l.toLst ::: (a :: r.toLst)
```
stew: the fact that you never use these last two parameters (height and size) in any pattern match ever should be a sign that they don't need to be part of the constructor.

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+
+}
+
+
+
+case class Branch[A] private[dogs](value: Option[A],
+                     left: BinaryTree[A],
+                     right: BinaryTree[A],
+                     private [dogs] val size: Int = 0,
+                     private[dogs] val height: Int = 0) extends BinaryTree[A] {
```
stew: size and height can always be calculated as left.size + right.size + 1 and left.height + right. height + 1, which means there is no reason to ask for the caller of the constructor to pass these values, the constructor itself can calculate them

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  private [dogs] override def balanceFactor: Int = this match {
+    case Branch(_, BTNil(), BTNil(), _, _)    =>  0
+    case Branch(_, BTNil(), r, _, _)          =>  r.height
+    case Branch(_, l, BTNil(), _, _)          =>  -1 * l.height
+    case Branch(_, l, r, _, _)                =>  r.height - l.height
+  }
+
+  private [dogs] override def update(): Branch[A] = this match {
```
stew: this method seems to only do one thing, which is to update the size and height members of a Branch method in case there is some situation where the size and height members were for some reason constructed with incorrect values?

Where is it possible that the size or height might possibly be incorrect that this might be anything but a no-op?

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+      case Branch(a, l, _, _, _)      =>  loop(l, a)
+    }
+
+    this match {
+      case BTNil()                    => None()
+      case Branch(a, l, _, _, _)      =>  loop(l, a)
+    }
+  }
+
+  def max(): Option[A] = {
```
stew: no reason to have a nullary parameter group here. This is traditionally used to indicate that a method is performing side effects, but this method does not.

So this should be:

    def max: Option[A]

not 

    def max(): Option[A]

core/src/main/scala/BinaryTree.scala
```diff
@@ -0,0 +1,250 @@
...
+        Branch(y.value, y.left, a, y.size, y.height).update()
+      }
+    }
+  }
+
+}
+
+
+
+case class Branch[A] private[dogs](value: Option[A],
```
stew: there is never a case that value is None here, so it doesn't make sense for this to be an Option.

tests/src/test/scala/BinaryTreeTest.scala
stew: also, you are only covering 3 of the billions of possible values which could be returned by compareTo.  The documentation for this method makes no guarantee that the values will be one of these three.

stew:
so after playing with this locally, I have found that there are just way too many problems. I"m closing this PR, I'm going to push up my refactoring of this branch directly instead with your commits in the history.

PR #11, opened 2016-02-05
Nocats

anicolaspp:
Discrete Interval Encoding Tree implementation.


tpolecat:
Looks like your branch is messed up ... this includes a lot of unrelated commits. I think you need to move this work to a new branch.

anicolaspp:
That's true. Seems I did something wrong with branches. I will be moving it to features/noncats

anicolaspp:
I did this wrong. I will pull request to the master

PR #12, opened 2016-02-05
Diev

anicolaspp:


core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+  def adj(x: A, y: A): Boolean = succ(x) == y
+}
+
+/**
+ * Represent a range [x, y] that can be generated using generate() or reverse().
+ */
+trait ARange[A] {
+  def apply(start: A, end: A): ARange[A]
+  def generate()(implicit discrete: Discrete[A]): List[A]
+  def reverse()(implicit discrete: Discrete[A]): List[A]
```
stew: what does this do? I'm guessing it does exactly what generate does but in reverse? is that right?

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+}
+
+/**
+ * Implementation of ARange that uses Discrete[A] to generate items in the range
+ */
+sealed class DRange[A] private [dogs] (val start: A, val end: A) extends ARange[A] {
```
stew: are there ever going to be other implementations of ARange besides this one? how would they differ?

tests/src/test/scala/dogs/DiscreteSpec.scala
```diff
@@ -0,0 +1,139 @@
...
+package dogs
+package tests
+
+import dogs.Diet.EmptyDiet
+import dogs.Order.{EQ, GT, LT, Ordering}
+import dogs.Predef._
+import org.scalacheck.Arbitrary.{arbitrary => getArbitrary}
+import org.scalatest.{FlatSpec, Matchers}
+
+class BigIntDiscrete extends Discrete[BigInt] {
```
stew: are we going to supply any of these? shouldn't we perhaps at least supply int?

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+import dogs._
+
+import scala.annotation.tailrec
+
+
+
+/**
+ * Represent discrete operations that can be performed on A
+ */
+trait Discrete[A] extends Order[A] {
```
stew: can we call this Enum?

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+  /**
+   * verify if x and y are consecutive
+   */
+  def adj(x: A, y: A): Boolean = succ(x) == y
+}
+
+/**
+ * Represent a range [x, y] that can be generated using generate() or reverse().
+ */
+trait ARange[A] {
```
stew: Is this part of the public interface? or is this an implementation detail?

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  override def apply(start: A, end: A): DRange[A] = DRange.apply(start, end)
+}
+
+object DRange {
+  def apply[A](x: A, y: A) = new DRange[A](x, y)
+}
+
+/**
```
stew: since this is the beginning of where we want someone reading, can we put it at the top of the file?

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+
+  val isEmpty: Boolean
+
+  /**
+   * verify x is a value is in the tree
+   */
+  def contains(x: A)(implicit discrete: Discrete[A]): Boolean = this match {
+    case EmptyDiet()          =>  false
+    case DietNode(a,b, l, r)  =>  (discrete.compare(x, a), discrete.compare(x, b)) match {
+      case (EQ, _)  =>  true
```
stew: I think we should just add "ge" and "le" methods to order to be able to ask about "greater than or equals" or "less than or equals"

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+      case (GT, LT) =>  true
+      case (LT, _)  =>  l.contains(x)
+      case (_, GT)  =>  r.contains(x)
+    }
+  }
+
+  /**
+   * return a list of all disjoint sets in the tree where each set is represented by ARange
+   */
+  def disjointSets(): List[DRange[A]] = this match {
```
stew: does this mean both DRange and ARange are part of the public interface?

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * add new value range [x, y] to de tree. If x > y then it will add range [y, x]
+   */
+  def add(x: A, y: A)(implicit discrete: Discrete[A]): Diet[A] = {
+    if (discrete.compare(x, y) == EQ) {
+      add(x)
+    }
+    else if (discrete.compare(x, y) == LT) {
+      add(x).add(discrete.succ(x), y)
```
stew: wait, we add each thing in the range individually? isn't this very inefficient? seems like adding a range to the Diet should be O(log n) not O(n log n)

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+            DietNode(x, y, l, r.add(value))
+      }
+      else d
+    }
+  }
+
+  /**
+   * remove x from the tree
+   */
+  def remove(x: A)(implicit discrete: Discrete[A]): Diet[A] = this match {
```
stew: we should also be able to remove a range

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,321 @@
...
+
+    def apply[A](): Diet[A] = this.asInstanceOf[Diet[A]]
+
+    override val isEmpty: Boolean = true
+  }
+
+  /**
+   * merge to Diets
+   */
+  private [dogs] def merge[A](l: Diet[A], r: Diet[A]): Diet[A] = (l, r) match {
```
stew: why private?

core/src/main/scala/Diet.scala
anicolaspp: Yes, you are correct.

core/src/main/scala/Diet.scala
anicolaspp: It only on implementation, I just didn't know how to treat it. Just it be only a class?

tests/src/test/scala/dogs/DiscreteSpec.scala
anicolaspp: I have one (BigInt) that I used for the test, I will be adding it here.

core/src/main/scala/Diet.scala
anicolaspp: Yes we can, I am fixing it now

core/src/main/scala/Diet.scala
anicolaspp: I think is should be part of the public interface. At this point we are not adding an ARange, but it can be implemented at some point. This comment also responds to one other other comments

core/src/main/scala/Diet.scala
anicolaspp: Yes we can, I will take care of it

core/src/main/scala/Diet.scala
anicolaspp: you are right, it should be List[ARanga[A]]

core/src/main/scala/Diet.scala
anicolaspp: The original paper doesn't talk about adding range, this addRange is very naive, it needs improvements, but not sure how to do it yet. We can remove it if you want to.

core/src/main/scala/Diet.scala
anicolaspp: The same problem, the original paper doesn't talk about remo

core/src/main/scala/Diet.scala
anicolaspp: No, it does not merge overlapping Diet(s). It is intended for remove. It is something to look at in the future.

core/src/main/scala/Diet.scala
anicolaspp: or we can add it later.

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,330 @@
...
+  def foldLeft[B](s: B)(f: (B, A) => B)(implicit discrete: Enum[A]): B = this match {
+    case EmptyDiet()          =>  s
+    case DietNode(a, b, l, r) =>  r.foldLeft(ARange(a, b).generate().foldLeft(l.foldLeft[B](s)(f))(f))(f)
+  }
+}
+
+object Diet {
+  def apply[A](): Diet[A] = empty()
+
+  def empty[A](): Diet[A] = EmptyDiet()
```
stew: since its not side-effecting, it shouldn't have a ()

core/src/main/scala/Diet.scala
```diff
@@ -0,0 +1,330 @@
...
+      case (GT, LT) =>  true
+      case (LT, _)  =>  l.contains(x)
+      case (_, GT)  =>  r.contains(x)
+    }
+  }
+
+  /**
+    * return a list of all disjoint sets in the tree where each set is represented by ARange
+    */
+  def disjointSets(): List[ARange[A]] = this match {
```
stew: not side-effecting, please remove the parameter list:

def disjointSets: List[ARange[A]]...

core/src/main/scala/Diet.scala
anicolaspp: I will do, it is just a habit from other languages.

core/src/main/scala/Diet.scala
anicolaspp: done too

PR #21, opened 2016-02-18
Diet benchmarks

anicolaspp:
I am completely new to JMH so please, review the small benchmarking code


tests/src/test/scala/DietSpec.scala
```diff
@@ -115,6 +115,13 @@ class DietTest extends FlatSpec with Matchers {
   import dogs.Predef._
   import Diet._
 
+  implicit object EnumInt extends Enum[Int] {
```
stew: we don't have to define this in the test, we have it in main:

https://github.com/stew/dogs/blob/master/core/src/main/scala/Enum.scala#L28

tests/src/test/scala/DietSpec.scala
```diff
@@ -239,7 +238,13 @@ class DietTest extends FlatSpec with Matchers {
 }
 
 class DietTestJoin extends FlatSpec with Matchers {
-  import Diet._
+
+  implicit object EnumInt extends Enum[Int] {
```
stew: we already have this in main

stew:
this needs to be rebased against master

anicolaspp:
I pull requested my master after rebase it. I think we can close this one now.

PR #23, opened 2016-02-22
Diet benchmarks - After rebase into master

anicolaspp:


bench/build.sbt.orig
```diff
@@ -0,0 +1,18 @@
...
+moduleName := "dogs-bench"
+
+publish := ()
+publishLocal := ()
+publishArtifact := false
+
+enablePlugins(JmhPlugin)
+
+libraryDependencies += "org.scalaz" %% "scalaz-core" % "7.2.0"
+<<<<<<< HEAD
```
fsvehla: Missed a merge conflict marker here

anicolaspp: for some reason it was present when it shouldn't. I just removed it

tests/src/test/scala/DietSpec.scala.orig
```diff
@@ -0,0 +1,369 @@
+package dogs
```
stew: this file shouldn't be here

PR #24, opened 2016-02-23
Diet page

anicolaspp:


stew:
wait, why are all those commits in this PR?

tests/src/test/scala/DietSpec.scala.orig
```diff
@@ -0,0 +1,369 @@
+package dogs
```
stew: this file shouldn't be here

stew:
Can we in general keep pull requests about one thing, This one is both adding benchmarks and documentation (which doesn't reference the benchmarks), and changing imports in a test...

PR #28, opened 2016-02-23
trying to fix it

anicolaspp:


bench/src/main/scala/DietAddBench.scala
```diff
@@ -0,0 +1,81 @@
...
+import dogs.Predef._
+import org.openjdk.jmh.annotations.{Benchmark, Scope, Setup, State}
+
+import scala.util.Random
+import scalaz.Diev
+
+@State(Scope.Benchmark)
+class DietAddBench {
+
+  implicit object scalazEnumInt extends scalaz.Enum[Int] {
```
stew: why define your own instead of using the one from scalaz?

https://github.com/scalaz/scalaz/blob/abbd0238972d71bd07724218d4f85341ab720389/core/src/main/scala/scalaz/std/AnyVal.scala#L234

bench/src/main/scala/DietBenchSearch.scala
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+  * Created by anicolaspp on 2/18/16.
+  */
+@State(Scope.Benchmark)
+class DietBenchSearch {
+
+  import dogs.Predef._
+
+  implicit object scalazEnumInt extends scalaz.Enum[Int] {
```
stew: don't define your own?

bench/src/main/scala/DietAddBench.scala
anicolaspp: I did not know we have one, I can fix it quickly

bench/src/main/scala/DietDataGen.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package dogs
+package bench
+
+import dogs.Predef._
+
+object DietDataGen {
+
+  def getWorstCaseData: scala.IndexedSeq[Int] = {
+    for (x <- scala.Range(1, 1000)
```
stew: fyi this is the same as Range(2,1000,2).to[IndexedSeq]

bench/src/main/scala/ListBench.scala
```diff
@@ -97,4 +97,4 @@ class ListMapperfy extends BigNumberLists {
   @Benchmark def scalazMapperfy(): Unit = {
     scalaz.map(_ + 1)
   }
-}
+}
```
stew: ?

bench/src/main/scala/DietBenchSearch.scala
anicolaspp: exactly, I will use the one in Scalaz

docs/src/main/tut/diet.md
```diff
@@ -0,0 +1,37 @@
...
+
+The discrete interval encoding tree is based on the observation that the set of integers { i| a<=i<=b } can be perfectly represented by the closed interval [a, b].
+
+`Diet` is a binary search tree where each node contains a set of continuos values. If a value is inserted and it fills the hole between to nodes (sets) then the two set become one since all value in the new set are consecutive.
+
+## Best and Worst Case Analysis.
+
+The best case is when there is no holes in the stored set, so the tree contains only one node (a single interval). In this case, it has ***O***(1) space requirement, inserting, deleting and finding a value is in constant time. 
+
+In the worst case scenario, there is a hole of size one (1) between each node and each node is in fact a set on size one (1). Under these circumstances, it behaves as a regular binary search tree.
```
stew: You mean linked list, not binary tree.

anicolaspp: I mean binary search tree

docs/src/main/tut/enum.md
```diff
@@ -0,0 +1,24 @@
...
+
+These operations are presented by the following functions.
+
+- `def succ(x)`:		Returns the successor value of `x`
+- `def pred(x)`: 	Returns the predecessor value of `x`
+- `def adj(i, j)`:	Returns if `i` and `j` are consecutives ( `succ(i) is j`)
+
+An example of the discrete operation on interger values could be: 
+
+```
```
stew: please make this "```tut" not just "```" so that it is compiled, this is the whole point of using [tut](https://github.com/tpolecat/tut)

docs/src/main/tut/diet.md
anicolaspp: as the original paper states

docs/src/main/tut/enum.md
anicolaspp: what symbol should I use?

anicolaspp: I think I get it now

docs/src/main/tut/diet.md
stew: if its not balanced, it is a linked list in the worst case, right? if you keep adding things in increasing order they keep getting added to the right most leaf

docs/src/main/tut/enum.md
stew: look at the other sources in this directory, and take a look at the docs for tut. the idea is that if you label these sections with tut, they will be compiled, and you see the output of a repl session in the resulting .md file. Compare the list.md to the resulting html page.

This way we make sure the code in our docs is always up to date

docs/src/main/tut/diet.md
anicolaspp: that is true, but binary search tree that is not balanced also has the same properties when adding in order values. However, if the values are not in increasing order, the same does not happens. That is why it is not a list

anicolaspp: Going back to another discussion we had before, if I had stored an AVL (with the worst case values of DIET) and then we load it into a binary search tree T, then T performance is not a O(n), right?

stew: This sentence is specifically talking about the worst case, where it is a list, not a tree

anicolaspp: That is false. Let's say that W is the set of values from the worst case scenario and have been sorted. We can select the item in the middle of W and add it to Diet. Then do the same to item to the left and to the items to the right until values from W have been added to Diet. Diet contains all worst case values and nothing else. However Diet is NOT a list.

Here is a link to the paper: https://web.engr.oregonstate.edu/~erwig/papers/Diet_JFP98.pdf

stew: if I insert 1 then 3 then 5 then 7 then 9, what will the tree look like?

will it look like:

      1
         \
           3
             \
               5
                 \
                   7
                     \
                       9

or will it be different?

PR #29, opened 2016-02-25
[WIP] Improve ListBench

mkhl:
This changeset is an attempt to improve the existing benchmarks a bit by
* introducing a parameter for the size of the lists we operate on
* avoiding looping the operation to measure inside a benchmarks,
  relying instead on the parameterized list sizes and JMHs builtin iterations


mkhl:
I'm not sure whether this change is desirable or what might be missing from it,
but I think it improves the status quo a slight bit,
and I imagine that measurements that back big O estimations will be useful in the long run.

stew:
@mkhl Thank you! This looks fantastic! I have no idea what I'm doing with jmh, so this is quite welcome, and please keep suggestions coming :)

PR #33, opened 2016-02-26
Def diet remove range fixes #31

### LINKED ISSUE #31: missing Diet.removeRange
we have a `def -` which does this, we just need the removeRange method for symmetry with addRange. When we do this, lets make sure to add it to the docs project

anicolaspp:


stew:
needs resync

PR #35, opened 2016-02-26
Range Docs and Inverted Range support

anicolaspp:


core/src/main/scala/Range.scala
```diff
@@ -110,15 +117,25 @@ sealed class Range[A](val start: A, val end: A) {
...
+        override def apply(l: A, r: A): Ordering = discrete.apply(l, r)
+      })
+  }
+
   private [dogs] def isEmpty: Boolean = false
 
   def apply(start: A, end: A): Range[A] = Range.apply(start, end)
 
-  override def toString: String = s"Range($start,$end)"
+  override def toString: String = if (isEmpty) s"[]" else s"[$start,$end]"
```
stew: Hmm, I kinda liked that the .toString of Range was something that I can paste back into the repl and get back the same range:

    scala> 1 to 10
    res0: dogs.Range[Int] = Range(1,10)

    scala> Range(1,10)
    res1: dogs.Range[Int] = Range(1,10)

PR #37, opened 2016-03-03
fix a typo in NotNull.scala

ritschwumm:


stew:
thank you very much, but we've removed NotNull, I'll double back and make sure this is fixed in cats where I was borrowing this from.

ritschwumm:
no problem - it's fixed in cats.

PR #44, opened 2016-03-04
There was a bug on addRange that allows us to add an empty range. Als…

anicolaspp:
…o, we should add a range [x, y] (x <= y) in increasing order even when [x, y] (x > y) is passed.


anicolaspp:
Once the two PRs get merged, I will remove a _.filter_ in dietshowable since now diet.addRange does not accepts rampty ranges

PR #49, opened 2016-03-28
Fibonnaci heap

anicolaspp:
I know we talked about heaps at the summit, but I needed the heap and noone has done it, so here it is.
Could you check it and merge it? 

I will be doing the doc and the benchmarks this week.

I also don't know if it makes sense to have fold and map on it????

### DIFF
--- a/core/src/main/scala/dogs/Heap.scala
+++ b/core/src/main/scala/dogs/Heap.scala
@@ -8,23 +8,48 @@ import algebra.Order
 import dogs.Predef._
 import dogs.Heap._
 
+
+
+/**
+ * Binary Heap
+ *
+ * Normally, binary heaps are not common the functional environments since they implementationis based on mutable
+ * arrays. However, this implementation is purely functional, based on the VLADIMIR KOSTYUKOV paper.
+ *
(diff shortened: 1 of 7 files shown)


core/src/main/scala/dogs/Heap.scala
```diff
@@ -0,0 +1,207 @@
...
+  def apply[A](x: A, l: Heap[A], r: Heap[A]): Heap[A] =
+    Branch(x, l, r, l.size + r.size + 1, scala.math.max(l.height, r.height) + 1)
+
+  private [dogs] case class Branch[A](min: A, left: Heap[A], right: Heap[A], size: Int, height: Int) extends Heap[A] {
+    override def isEmpty: Boolean = false
+
+    override def getMin: Option[A] = Some(min)
+  }
+
+  private [dogs] case object Leaf extends Heap[Option[Nothing]] {
```
stew: why is this extending Heap[Option[Nothing]] instead of Heap[Nothing] ?

anicolaspp: min in Leaf is a nothing. How do I return Nothing? 

val min: Nothing = ????  <- what to put in here

PR #53, opened 2016-04-10
Sorted

anicolaspp:
We should be able to sort List[A]. By using type classes we can extend it so we can use the sorting algorithm we want. I provided quicksort and and heapsort (based on the heap I created before)


core/src/main/scala/dogs/BiPartition.scala
```diff
@@ -0,0 +1,36 @@
...
+import simulacrum.typeclass
+import scala.{inline,Iterable}
+import java.lang.{String,StringBuilder}
+import scala.annotation.{tailrec}
+import dogs.syntax.birds._
+import cats._
+
+
+@typeclass trait BiPartition[A] {
+  def partition(predicate: A => Boolean): (List[A], List[A])
```
stew: Hmm, this really seems like it should be able to be derived from some other typeclass we already have in cats. It looks a lot like separate, can we generalize separate to fit this as well? Why do we want List as the result? Can we generalize the result to a Alternative? I'm guessing the performance loss there would suck

core/src/main/scala/dogs/Heap.scala
```diff
@@ -67,21 +67,21 @@ sealed abstract class Heap[A] {
...
+//    def loop(i: Int, xs: scala.List[A]): Heap[A] =
+//      if (i < xs.length) {
+//        bubbleDown(xs(i), loop(2 * i + 1, xs), loop(2 * i + 2, xs))
+//      }
+//      else {
+//        Leaf()
+//      }
+//
+//    loop(0, a.toScalaList)
+//  }
```
stew: seems like we don't need this ^

core/src/main/scala/dogs/Sorted.scala
```diff
@@ -0,0 +1,51 @@
...
+import simulacrum.typeclass
+import scala.{inline,Iterable}
+import java.lang.{String,StringBuilder}
+import scala.annotation.{tailrec}
+import dogs.syntax.birds._
+import cats._
+
+
+@typeclass trait Sorted[A] {
+  def sorted(xs: List[A]): List[A]
```
stew: shouldn't only be list

core/src/main/scala/dogs/BiPartition.scala
anicolaspp: how we could do this then?

core/src/main/scala/dogs/Sorted.scala
anicolaspp: how do we generalize xs? the result should still be a List[A], right?

core/src/main/scala/dogs/BiPartition.scala
anicolaspp: how does Alternative work?

core/src/main/scala/dogs/Sorted.scala
anicolaspp: is this the right way to define it? 

@typeclass trait Sorted[M[_]] {
  def sorted[A](xs: M[A]): List[A]
}

tests/src/test/scala/ListSpec.scala
```diff
@@ -2,15 +2,23 @@ package dogs
 package tests
 
 import Predef._
+import com.oracle.webservices.internal.api.message.PropertySet.Property
```
stew: ??

core/src/main/scala/dogs/BiPartition.scala
stew: Alternative is the thing that is both a Applicative (giving us pure, a way to go from A => F[A]), and MonoidK, which is a higher kinded monoid, giving us empty: () => F[A] and append: F[A] => F[A] => F[A].  This is the minimum you need from the cats typeclasses to get something like a builder. Because from these three functions. for example, you can do:

    def prepend[F[_]](fa: F[A])(a: A)(implicit F: Alternative[F]) = F.append(F.pure(a), fa)
    def append[F[_]](fa: F[A])(a: A)(implicit F: Alternative[F]) = F.append(fa, F.pure(a))

stew: But I guess in general this typeclass doesn't make sense to me, It says "for some type A, I can take a function A => Boolean, and produce from that two Lists, but it doesn't say where the source of these values is? Why is is parameterized by A?

Is there some value of this typeclass outside of quicksort?

anicolaspp: It is true, it should be able to get an structure F[A] and transform it into a Partition[A] where BiPartition[A] is an instance of Partition[A].

Having a way to partition structures is important, for instance, in Diet, when you say to diet.intervals that is a partition of Diet. 

If you have a Set you could apply a function to it and get some kind of partition out of it (equivalence classes). 

You could partition a tree in sub-trees, or a graph, in subgraphs. 

I think it would be nice to have, I just need to change its definition.

PR #55, opened 2016-04-19
Topic/build1

InTheNow:


InTheNow:
Also see https://github.com/scalatest/scalatest/issues/661 for a description of the settings and note that they will change again, soon.

PR #59, opened 2016-06-26
Optimizing Range Operations

anicolaspp:


anicolaspp:
The build fail because there are tests that take TOO long

anicolaspp:
I re-submitted the changes so this is not needed any more.

PR #62, opened 2016-10-13
upgrade cats library to 0.7.2

stew:


arosien:
Any ideas on the test failures? Looks like some kind of hang.

Nudging because upgrading the cats dep would be awesome.

stew:
@arosien No, I'm stumped. It's intermittent, sometimes the build just hangs, when it does, it is AFTER I've seen ALL of the expected tests output success, then output just stops coming :(

BennyHill:
@stew There were, I think, similar issues in cats with the scalacheck generator that finally went away with an upgrade of scalacheck to 1.13.x. That upgrade also needed changes in the `Laws` to incorporate `Cogen`, etc. 

So perhaps even if you don't  technically _need_ to upgrade scalacheck, it might be worth doing so - in which case https://github.com/stew/dogs/issues/13 may as well be done as well.

I'd be happy to lend a hand here, esp if you could provide some blueprints for the changes

BennyHill:
OK, having said all that, I think I've got a build working _without_ having to upgrade everything else for now - will PR shortly

PR #71, opened 2016-11-05
Optimizations on Range

anicolaspp:
I re-sync the master and added everything in one commit


anicolaspp:
This fails because it takes too much time to build

stew:
I don't understand the point of memoizing anything here. This memoized value is only used in a single method, right?

anicolaspp:
No really. In Range, everything goes down to generate the range if the values are needed. With this change, the values in the range are only generated once. The following request to the values within the Range will returned the generated ones.

PR #73, opened 2016-11-12
FlatMap instance for Map

gyfarkas:
Hi, I've created FlatMap instance for Map[K,V]. 
I had to rewrite flatMap method, I took inspiration from cats' Map instances. 

Thanks, 
Gy


stew:
👍 Thanks!

PR #81, opened 2017-03-06
Change CrossVersion.full to CrossVersion.patch

rilut:


stew:
thanks! I got two such patches though and merged the other :)

PR #83, opened 2017-03-07
Feature/disjoint set

lewismj:
Hi,

This is just a basic implementation of a Disjoint-set. It is probably a structure that will be useful for other data-structures rather than something used by itself.

I've tried not to over elaborate and add 'too many' (i.e. graph specific) methods to the interface and keep this as simple as possible, at least for now. 

thanks,
Mike

### DIFF
--- a/core/src/main/scala/dogs/DisjointSet.scala
+++ b/core/src/main/scala/dogs/DisjointSet.scala
@@ -1,0 +1,97 @@
+package dogs
+
+import Predef._
+import cats._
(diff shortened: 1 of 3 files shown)


lewismj:
Will try re-running, 'sbt test' runs fine on my local machine, but Travis seemed to have hung up for some reason (think after the additional tests I added were run).

anicolaspp:
we have this problem for long time now

PR #86, opened 2017-03-09
Disjoint-set implementation.

lewismj:
This PR implements a DisjointSet. I've tied up the change to a single commit. Includes the test and doc.

### DIFF
--- a/core/src/main/scala/dogs/DisjointSet.scala
+++ b/core/src/main/scala/dogs/DisjointSet.scala
@@ -1,0 +1,91 @@
+package dogs
+
+import Predef._
+import cats._
+import cats.instances.all._
+import scala.annotation.tailrec
+
+
+/**
+  * A disjoint set is a data structure that keeps track of a set of
+  * elements that are partitioned into a number of disjoint subsets.
+  * Useful for finding the number of components within a graph.
+  */
+final class DisjointSet[T] private (parents: Map[T,T], ranks: Map[T,Int], nComponents: Int) {
+
+  /**
+    * Return the number of components.
+    */
+  def components: Int = nComponents
+
+  /**
+    * Finds the root parent of a given element.
+    */
+  @tailrec
+  def find(u: T)(implicit order: Order[T]): Option[T] = {
+    parents.get(u) match {
+      case None() => None()
+      case Some(v) => if (u == v) Some(v) else find(v)
+    }
+  }
+
+  /**
+    * Given two elements u and v, return Some(find(u),find(v))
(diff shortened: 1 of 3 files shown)


lewismj:
Not sure if using ._1, _.2 is being a bit lazy in my pattern matching, maybe better to write Some ((xr,yr)), but good to know if the basic structure looks ok and if so, I'll tidy that bit up.

pfcoperez:
Hi @lewismj , I was about to open the a PR with the functionality for disjoint sets when I found yours. I wanted to bring into dogs the implementation I published at 
https://github.com/pfcoperez/algorithmaday/tree/master/src/main/scala/org/pfcoperez/dailyalgorithm/datastructures/sets

Which is a pure functional version of the structure described at [CLRS](https://en.wikipedia.org/wiki/Introduction_to_Algorithms). I implemented it in Python in 2012 and [provided an small summary of the structure described by Thomas Cormen](http://orionsword.no-ip.org/blog/wordpress/?p=246) and I adapted it to a functional style. 

In fact I also provided a state monad set of operation thus making it able to be used using `for` notation in such a way that the state propagation is transparent for the user:

Use of the sate monad

The point is that I was reviewing your implementation and I see its simple and nice and it also implements **union by rank heuristic**. However it seems to lack the **path compression heuristic**. That is OK for most cases but it can lead to degraded states  where lookups can be as slow as **O(n)**. Check the how the tree depth increases in this example where the path compression is not being applied:

Disjoint sets degradation

I hope you don't mind I open a PR with my version so we can discuss the differences.

lewismj:
Hi @pfcoperez If the dogs maintianers don't mind cats dep, I would raise a PR for your version.
I think there is a 'professional' graph lib called quiver, not sure if there is a preference to put this sort of stuff into that. I've been experimenting with some old graph stuff that I had in C++ (well it was C++ without the STL for speed) and wanted to try a Scala version. It is years old, some stuff I had for algos.
If you don't mind, I might try to put your implementation into the 'experimental/learning' lib that I have in my repo, I'd like to take a look further.
I think initially, I was in two minds, about the path compression, as I thought the implementation without it was easy to read etc.. but its probably better with your stuff in it.

pfcoperez:
@lewismj They are already using Cats for their data structures at Dogs, for example, they use `Show` type class at `Heap`. 

In fact, I've changed the original implementation from my "algorithmaday" repo in order to implement union using the find `State` operation.

I didn't know `quiver`, thanks for the tip.

Of course you can include my code in your library, it is open source, use it as you wish :wink:

PR #93, opened 2017-06-13
Added emacs ensime cache directory to .gitignore file

pfcoperez:
When Emacs is used along side of Ensime, it creates a local directory  `.ensime_cache` which is not listed in this project `.gitignore` file thus showing changes to stage. 

Ths PR adds that entry to the git ignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ project/boot
 target
 .ensime
 .ensime_lucene
+.ensime_cache
 TAGS
 \#*#
 *~


fsvehla:
Not an Dogs dev, but you might be interested to set up a global gitignore for .ensime_cache / .proctile, etc

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

pfcoperez:
@fsvehla Thanks for your comment, that would solve the problem in my machine for sure. I opened the PR since there are some other entries masking ensime in the current `.gitignore`. By doing it in this blacklist you might avoid accidents in machines where this global setting has not been enabled by the user.

sarahgerweck:
This is a matter of opinion, but I agree with @pfcoperez that it's a good idea for OSS Scala projects to have a `.gitignore` that includes the artifacts created by the most common editors & tools like SBT, Ensime, Vim, Emacs, VS Code, IntelliJ and Eclipse. My personal rule is that the project should make the common cases work by default, but the user should be responsible for customizing their environment if they're doing something weird. This is all opinion though.

Issue #66, opened 2016-10-26
DietMap


velvia:
I found Diet, which seems quite nice, but I'm particularly interested in a DietMap, which would allow you to store values associated with each range.

    dietMap.addRange((1, 10), myCaseClass)
    def intersection(range): Seq[V]

anicolaspp:
I could add dietmap[A,B] to store items of type B associated with ranges of type A. I dont think it is a problem. I just need other things committed first.

anicolaspp:
@stew I am going to start working on this. The interesting part is merging two nodes. Each Node needs the `Range` and and elements in the range. I am thinking on using shapeless to do this since we don't need to fix a type `A`. A `Node` could potentially look like this: 

```
case class Node[A <: HList, B](range: Range[B], item: A)
```
in other words, we could do something like this:
```
Node(Range(10, 20), 5 :: HNil) + 
Node(Range(8, 18, "Hello" :: HNil)
```
will result in:
```
Node(Range(9, 20), 5 :: "Hello" :: HNil)
```

What do you think?

stew:
I think that it should require a Semigroup on A instead of using an HList

anicolaspp:
That will work too. I will look into it

anicolaspp:
@velvia I have done this and open #94 make sure you look at the commit message to see how to use it.

velvia:
Thanks very much!

anicolaspp:
@larsrh I am back to write some code for this project. I would like to resubmit some of my old PRs. between @stew and I wrote most of the structures here and I would like to continue with it.

larsrh:
That's great! I'd be happy to take a look at them.

PR #94, opened 2017-06-21
1- `VRange[A, B]` has been added as extension of Range such as it hol…

anicolaspp:
…ds the same properties but also has an additional value.

2- `Diet` should be able to add VRange since a VRange is actually a Range
3- `Diet.focus` returns the range passed in if it exists as a complete node. Then one could pattern match on it to extract the associated value if it is a VRange.

#66


larsrh:
As you may have noticed, the project has been moved to the Typelevel organization and the maintainer team has changed. I apologize for this PR sitting for so long. I'm closing it because it's outdated. If you'd like to restart work on this, please feel free to reopen and/or ping us on [Gitter](https://gitter.im/typelevel/cats).

PR #95, opened 2017-08-08
cats 1.0.0-MF

tpolecat:
It would be nice to get a milestone out at some point.


tpolecat:
Hey I'm a maintainer. I'll be damned.

tpolecat:
Hm looks like CI is trying to build JS but it's not actually set up in `build.sbt`.

anicolaspp:
What is going on with @stew? Where is he?

tpolecat:
I'm trying to get scala-js working. PR coming.

tpolecat:
ok attempted to fix the build … it's tied up with the mods here because I needed `Is` to get rid of a compiler warning

tpolecat:
son of bitch. i think it's good but something in the build is randomly hanging

tpolecat:
Ok at long last this builds. @stew could you take a look? Due to build changes I think this needs to go to the head of the line. I recommend a squash commit because there's a bunch of bullshit in the history.

PR #96, opened 2017-08-14
attempt to fix sjs build

tpolecat:
let's see how this goes


tpolecat:
hm something's taking too long

tpolecat:
Hm it wasn't building for 2.12 either.

tpolecat:
moving this work over to the cats PR

PR #113, opened 2018-09-20
Move Files Into Cats Collections Package

ChristopherDavenport:
Just Moves Files So That They are in `cats.collections`


SethTisue:
Scala uses `scala.collection`, singular. Should this be `cats.collection`, singular?

Issue #105, opened 2018-09-19
complete migration


kailuowang:
- [ ] move root package to `cats.collections`
- [ ] s/dogs/cats-collections/ in build and documentation
- [ ] we might not want to start cats-collections with all existing dogs collections. e.g. `DList` (now we have `Chain` in cats), `Streaming` which we have more mature solution in the ecosystem , etc. cc @mpilquist. We can create a `legacy` package or module and move those that we are no so sure in. 
- [ ] make sure the new project website works

danslapman:
`birds` can be moved into mouse - `mouse.any` already have thrush

larsrh:
I'm assigning myself for now. I'll work on at least the website issue this weekend (together with an overhaul of how the typelevel.org website works).

larsrh:
* [ ] check stale branches

Any takers? I've already deleted those that have been merged, but there are a few which are ahead of `master`.

kailuowang:
@larsrh I reviewed. Those are indeed outdated and safe to be deleted. 
Another thing is that I didn't receive the Email invitation to be a collaborator of the project yet. Maybe it's github?

larsrh:
@kailuowang I just forgot to do it. Done now.

larsrh:
I've created a 0.7 and a Backlog milestone. Everyone, it would be good if you could go through open PRs/issues and mark them accordingly.

kailuowang:
I went through the issues PRs, all seem pretty stale. Maybe we can just close them all and let the stakeholders reopen it if they choose to re-engage?

larsrh:
> Maybe we can just close them all and let the stakeholders reopen it if they choose to re-engage?

Done.

PR #116, opened 2018-09-24
[WIP] Deprecations & removals

larsrh:
According to [@mpilquist's comment](https://github.com/typelevel/cats-collections/issues/105#issuecomment-422614258) in #105.

Still a work in progress.

Instead of removing the things altogether, I think we should go through a deprecation cycle. This also has a practical advantage: `Streaming` is used in the implementation of some efficient algorithms in this library. By deprecating instead of removing we don't have to burden ourselves with rewriting those algorithms. This can happen in a second step after release of `0.7`.


core/src/main/scala/cats/collections/DList.scala
```diff
@@ -6,30 +6,35 @@ import cats._, cats.implicits._
...
  * Based on `Data.DList`, a Haskell library by Don Stewart.
  */
+@deprecated("DLists are superseded by cats.Chain", "cats-collections 0.7.0")
 final class DList[A](val run: List[A] => Eval[List[A]]) {
   import Eval._
 
   /**
    * Prepend the contents of this dlist to the given tail.
    */
+  @deprecated("DLists are superseded by cats.Chain", "cats-collections 0.7.0")
```
LukaJCB: Does it make sense to point to `cats.data.Chain` instead? :)

larsrh:
Missing:

* [ ] `Range`
* [ ] `Enum`

Commence bikeshedding!

larsrh:
Proposal: Keep `Range` as is. Rename `Enum` to `Discrete`.

larsrh:
I consider this no longer WIP. Ready for review!

dwijnand:
JFYI the docs still say Set: https://github.com/typelevel/cats-collections/blob/master/docs/src/main/tut/set.md

larsrh:
Yeah, also the source references are wrong.

PR #117, opened 2018-09-25
Fix the build

larsrh:


build.sbt
```diff
@@ -75,8 +75,8 @@ lazy val commonSettings =
     fork in test := true
   )
 
-addCommandAlias("build", ";core/compile;core/test;tests/test")
-addCommandAlias("validate", ";scalastyle;build")
+addCommandAlias("build", ";compile;test")
```
larsrh: Someone who understands scoverage, please check that this change is fine. (There are no tests in `core` though.)

larsrh:
Okay, that didn't work out. Paging the coverage experts. @kailuowang

kailuowang:
doesn't seem to be a scoverage issue. It seems that `bench` doesn't compile on `2.11.12` , I added the buildSettings to it. Seems fixed the problem locally, although I don't know the root cause of jmb freaking out when there is no sepcific scalaVersion / crossScalaVersion set. That error message is puzzling

PR #121, opened 2018-09-26
Deletions

larsrh:
This dismantles `Streaming` as far as possible, but we still have to implement some missing things here and there.


adelbertc:
Oh noes we don't like `Streaming`? I've actually found myself wanting it occasionally when I want a lazy list but no `IO`.

larsrh:
> Oh noes we don't like `Streaming`? I've actually found myself wanting it occasionally when I want a lazy list but no `IO`.

It feels like this is not the appropriate place for this type. fs2 has it, even for the pure case.

larsrh:
@adelbertc see also #105 where @mpilquist proposed removal.

adelbertc:
> fs2 has it, even for the pure case.

fair enough

larsrh:
> fair enough

I'll construe that as your approval :stuck_out_tongue_winking_eye:

Issue #120, opened 2018-09-26
Eq[AvlSet[A]]


larsrh:
Curiously, there appear to be no instances for `AvlSet`.

larsrh:
If we can implement this without `Streaming`, we can also get the `Eq` instance for `AvlMap` for free.

kailuowang:
closed by #124

PR #124, opened 2018-09-27
Set equality

larsrh:
This reworks set equality to use `Iterator` internally.

I noticed a few other oddities. For example, it appears that the `Arbitrary` instances aren't actually used much in tests.

~(Relatedly, those instances should probably be published.)~ _Update:_ they are published now


larsrh:
Closes #120.

Issue #122, opened 2018-09-26
Remove all deprecated methods and classes


larsrh:
Follow-up to #105 and #116.

PR #129, opened 2018-10-03
remove Streaming from Dequeue

larsrh:
Towards #122.

### DIFF
--- a/core/src/main/scala/cats/collections/Dequeue.scala
+++ b/core/src/main/scala/cats/collections/Dequeue.scala
@@ -1,7 +1,11 @@
 package cats.collections
 
-import cats._, cats.data.NonEmptyList
+import cats._
+import cats.data.NonEmptyList
+
+import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/collections/Dequeue.scala
```diff
@@ -87,11 +91,32 @@ sealed abstract class Dequeue[A] {
...
+    }
+  }
+
+  def to[Col[_]](implicit cbf: CanBuildFrom[Nothing, A, Col[A]]): Col[A] = {
+    val builder = cbf()
+    @tailrec def go(cur: Dequeue[A]): Unit = cur.uncons match {
+      case Some((a, rest)) =>
+        builder += a
+        go(rest)
+      case _ =>
```
ChristopherDavenport: should this have an explicit `()`?

larsrh: No, I don't think so. Like an `if` without an `else`, this is implicitly unit.

PR #130, opened 2018-10-03
Scala 2.13.0-M4

larsrh:
Add cross-compilation to M4.

Remaining problems:
* [x] scala.js linking error in tests (scala-js/scala-js#3445)
* [x] JVM test failure in M4 (probably an equality issue)
* [x] scoverage unavailable (#131)

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,4 +8,4 @@ addSbtPlugin("org.scoverage"      % "sbt-scoverage"          % "1.5.1")
 addSbtPlugin("org.xerial.sbt"     % "sbt-sonatype"           % "2.3")
 
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject"      % "0.6.0")
-addSbtPlugin("org.scala-js"       % "sbt-scalajs"                   % "0.6.25")
+addSbtPlugin("org.scala-js"       % "sbt-scalajs"                   % "0.6.24")
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ sudo: false
 scala:
 - 2.11.12
 - 2.12.6
+- 2.13.0-M4
 sbt_args: -J-Xss4m -J-Xmx6G -J-Xms1G
 jdk:
 - oraclejdk8
--- a/bench/src/main/scala/bignumbers.scala
+++ b/bench/src/main/scala/bignumbers.scala
@@ -14,7 +14,7 @@ trait BigNumberLists {
 
   @Setup
   def setup: Unit = {
-    scala = (1 to n).to[List]
+    scala = (1 to n).toList
     scalazlst = IList((1 to n):_*)
   }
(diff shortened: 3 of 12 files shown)


jozic:
isn't M5 already out?

SethTisue:
it is, but various dependencies aren't published for M5 yet.

it makes sense to at least make a stab at M4 support as a stepping stone to M5. if the effort runs in real trouble, it might ultimately make sense to give up and do M5 instead. but M4 and M5 are very similar so the partial work on M4 won't be wasted.

larsrh:
This turned out to be much more complicated than expected. It is really hard to convince sbt to _not_ compile some submodules on a particular Scala version. I tried various incantations to persuade sbt to set empty source directories for the JS tests, but to no avail. I'm open to suggestions.

larsrh:
Or I could just downgrade scala.js 🤷 This seems to work.

larsrh:
Huzzah! All green!

larsrh:
@SethTisue I'd be interested in your opinion. I know that a compatibility library for 2.12/2.13 collections exists, but it seemed to be overkill.

larsrh:
@ChristopherDavenport There's [a compatibility library](https://github.com/scala/scala-collection-compat/). I was considering pulling that in as a dependency, but I decided that the cost of the dependency outweighs the cost of adding a small fragment of it here. But I can be persuaded either way.

SethTisue:
I agree that scala-collection-compat shouldn't be pulled in unless the amount of effort and/or code it saves is considerable. Dependencies are the devil. I think it's normal to do as you've done here with `Factory`.

project/plugins.sbt
```diff
@@ -8,4 +8,4 @@ addSbtPlugin("org.scoverage"      % "sbt-scoverage"          % "1.5.1")
 addSbtPlugin("org.xerial.sbt"     % "sbt-sonatype"           % "2.3")
 
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject"      % "0.6.0")
-addSbtPlugin("org.scala-js"       % "sbt-scalajs"                   % "0.6.25")
+addSbtPlugin("org.scala-js"       % "sbt-scalajs"                   % "0.6.24")
```
ChristopherDavenport: Why the downgrade here?

SethTisue: apparently 0.6.25 works with Scala 2.13.0-M5 but not 2.13.0-M4, see 
https://github.com/typelevel/cats-collections/pull/130#issuecomment-427152636

it's plausible, there's no real reason for Seb to support multiple 2.13 milestones in the same version

larsrh: Yes, there's indeed a bug that prevents Scalacheck from being used with M4. But our upstream dependencies are not ready for M5 yet. So that's the only compromise I could come up with.

ChristopherDavenport: 👍

PR #138, opened 2018-10-12
Change `Discrete` instance to generic

lJoublanc:
This is to support `Long` instances, but will also cover anything that has a `scala.Integral` instance. The `specialized` annotation means there should be no performance loss, but I haven't benchmarked.


larsrh:
@johnynek While that's true, I don't think this is a concern for cats-collections, which is in pre-1.x, no?

larsrh:
Thank you for your contribution!

Issue #149, opened 2018-12-15
Adding into Diets doesn't work


larsrh:
```scala
scala> val d = Diet.empty[Int] + Range(572, 573) + 567 + 570 + Range(569, 574)
scala> d.toList
res30: List[Int] = List(567, 570, 569, 570, 571, 572, 573, 574)
```

anicolaspp:
the `.toList` out does not look right. It should not have repeated values.

larsrh:
Indeed. #150 should fix it, please have a look.

Issue #147, opened 2018-12-15
Removing from Diets doesn't work


larsrh:
```scala
scala> Diet.empty[Int].addRange(Range(1,2)).addRange(Range(4,5)).removeRange(Range(0,6))
res4: cats.collections.Diet[Int] = DietNode(Range(4,5),EmptyDiet,EmptyDiet)
```

Correct result should be an empty diet.

PR #150, opened 2018-12-15
Fixes in Diet and tests

larsrh:
I'm not 100% convinced that my changes are accurate. But at least #147 and #149 seem to be fixed.

### DIFF
--- a/core/src/main/scala/cats/collections/Diet.scala
+++ b/core/src/main/scala/cats/collections/Diet.scala
@@ -53,11 +53,13 @@ sealed abstract class Diet[A] {
   // a new copy of our tree which contains no ranges greater than `a`
   // and a new lower bound for our parent range (in case a was
   // adjacent to our greatest Range)
-  @tailrec
   private def noMoreThan(a: A)(implicit order: Order[A], enum: Discrete[A]): (Diet[A], A) =
     this match {
-      case DietNode(rng,l,_) =>
-        if(order.gt(a, rng.end)) (this, a)
+      case DietNode(rng,l,r) =>
+        if(order.gt(a, rng.end)) {
+          val (r2, a2) = r.noMoreThan(a)
+          (DietNode(rng,l,r2), order.min(a, a2))
+        }
         else if(order.gteqv(a, rng.start)) (l, rng.start)
         else l.noMoreThan(a)
       case x => (x,a)
@@ -68,11 +70,13 @@ sealed abstract class Diet[A] {
   // new copy of our tree which contains no ranges less than `a` and a
   // new upper bound for our parent range (in case a was adjacent to
   // our leaast Range)
-  @tailrec
   private def noLessThan(a: A)(implicit order: Order[A], enum: Discrete[A]): (Diet[A], A) =
     this match {
-      case DietNode(rng,_,r) =>
-        if(order.lt(a, rng.start)) (this, a)
+      case DietNode(rng,l,r) =>
+        if(order.lt(a, rng.start)) {
+          val (l2, a2) = l.noLessThan(a)
+          (DietNode(rng,l2,r), order.max(a, a2))
+        }
         else if(order.lteqv(a, rng.end)) (r, rng.end)
         else r.noLessThan(a)
       case x => (x,a)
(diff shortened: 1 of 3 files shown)


johnynek:
@larsrh by the way, if you want to merge this, I am willing to say +1 and I can follow up by adding a bunch of the tests that I have suggested if you don't have time to get to it. Collections are some of the most testable examples of code, but also somewhat subtle (as we have seen here with a bug), so I'd love to shoot for 100% coverage.

larsrh:
Your comments are all good (I knew it would be worthwhile to ask you for a review!).

larsrh:
@johnynek I have implemented most of your suggestions. I can live without coverage of a function that's just an alias, though.

OTOH, I have deprecated `map`, because upon closer inspection, its semantics is entirely unclear. The desired list homomorphism property can't hold for arbitrary functions. For example, let's say you have a `Range(1,2)` and you map it with `_*2`. That would give you `Range(2,4)`. But `List(2,3,4)` is not equal to `List(2,4)` (which the law would require). The current implementation seems broken and not worth fixing.

core/src/main/scala/cats/collections/Diet.scala
```diff
@@ -141,21 +144,27 @@ sealed abstract class Diet[A] {
...
   }
 
+  /**
+    * remove a range from Diet
+    */
+  def removeRange(range: Range[A])(implicit enum: Discrete[A], order: Order[A]): Diet[A] =
+    if (order.lteqv(range.start, range.end))
+      removeRangeIncreasing(range)
+    else
+      removeRangeIncreasing(range.reverse)
```
johnynek: This is surprising to me. I would have thought a range with end < start would be an empty Range. Is this not the case?

tests/src/test/scala/cats/collections/DietSpec.scala
```diff
@@ -1,375 +1,231 @@
...
-  test("add disjoint range"){
+  test("shown empty"){
     val diet = Diet.empty[Int]
 
-    val result = diet.addRange(Range(0, 100))
-
-    val other = result.intervals.map(l => l.generate)
-
-    other should matchTo(List(Range(0, 100).toList))
+    diet.show should be ("{ }")
```
johnynek: Personally I’d like the string Diet to appear somewhere here

tests/src/test/scala/cats/collections/DietSpec.scala
```diff
@@ -1,375 +1,231 @@
...
+  test("merge")(forAll { (rs1: Ranges, rs2: Ranges) =>
+    val diet1 = rs1.toDiet
+    val diet2 = rs2.toDiet
+    (diet1 ++ diet2).toList should be((rs1.toSet ++ rs2.toSet).toList.sorted)
+  })
 
-  test("intersect with another diet"){
-    val diet = Diet.empty[Int] + Range(20, 30)
+  test("intersection range")(forAll { (rs: Ranges, m: Int, n: Int) =>
+    whenever(m > n) {
```
johnynek: I think this can be >=

core/src/main/scala/cats/collections/Diet.scala
johnynek: I guess this is a property of Range I didn’t know. I guess there are some reasons for it (maybe it makes map on Range more useful, or maybe just more confusing, not sure which).

tests/src/test/scala/cats/collections/DietSpec.scala
larsrh: This is a separate discussion, I think. I'll open a ticket.

core/src/main/scala/cats/collections/Diet.scala
larsrh: The `Range` semantics admits `start > end` and then walks backwards (see the sources). The tests produce such ranges and apparently it works everywhere.

tests/src/test/scala/cats/collections/DietSpec.scala
```diff
@@ -1,375 +1,231 @@
...
+      // there may be structural changes, so fall back to list comparison
+      d.add(elem).toList should be(d.toList)
+    )
+  })
 
-    diet.toList() should matchTo(List(9, 10, 12, 13, 14, 20, 21, 23, 24))
+  def invariant[A](d: Diet[A])(implicit order: Order[A], enum: Discrete[A]): Boolean = d match {
+    case Diet.DietNode(rng, left, right) =>
+      order.lteqv(rng.start, rng.end) &&
+        left.toList.forall(order.lt(_, rng.start)) &&
```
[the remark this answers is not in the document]
larsrh: Right, it could be tighter. I suppose I would need to compare the predecessor of rng.start.

Issue #151, opened 2018-12-27
Would you accept an immutable list implementation of Okasaki


johnynek:
see:

https://gist.github.com/johnynek/5f4f97772595e2f65092cf51d186d6f2

it's a lot of work for everyone to send a complete PR. I want to get an idea if this is interesting to the maintainers before I do.

Alternatively, anyone is free to copy all or any of my code to send an implementation.

larsrh:
Pending some tests and a more thorough review, I'd be in favour of inclusion. I appreciate your work.

larsrh:
Merged.

johnynek:
Thank you!

PR #152, opened 2018-12-28
Add TreeList, an Okasaki immutable list with fast indexing

johnynek:
following up on #151 

I think the test coverage is complete (hopefully we will see in the report on the PR).

We can discuss the implementation details, but I think it is in pretty good shape.

Thanks for looking.

cc @larsrh

### DIFF
--- a/core/src/main/scala/cats/collections/TreeList.scala
+++ b/core/src/main/scala/cats/collections/TreeList.scala
@@ -1,6 +1,6 @@
 package cats.collections
 
-import cats.{Applicative, Alternative, CoflatMap, Eval, Eq, Monad, Monoid, Order, PartialOrder, Semigroup, Show, Traverse}
+import cats.{Applicative, Alternative, CoflatMap, Eval, Eq, Functor, FunctorFilter, Monad, Monoid, Order, PartialOrder, Semigroup, Show, Traverse}
 
 import cats.implicits._
 import scala.annotation.tailrec
@@ -11,7 +11,9 @@ import scala.annotation.tailrec
  *
  * A consequence of the log N complexity is that naive recursion on the inner
  * methods will (almost) never blow the stack since the depth of the structure
- * is log N, this greatly simplifies many methods
+ * is log N, this greatly simplifies many methods. A key example is that unlike
+ * List, using a TreeList you can sequence and traverse very large lists without
+ * blowing the stack since the stack depth is only log N.
  *
  * This data-structure is useful when you want fast cons and uncons, but also
  * want to index. It does not have an optimized concatenation. It can iterate
@@ -80,6 +82,14 @@ sealed abstract class TreeList[+A] {
    * access using get
    */
   def toReverseIterator: Iterator[A]
+  /**
+   * If the given index is in the list, update it, else
+   * return the current list with no change.
+   *
+   * O(log N)
+   */
+  def updatedOrThis[A1 >: A](idx: Long, value: A1): TreeList[A1]
+
   /**
    * map to a type with a Monoid and combine in the order of the
    * list
@@ -151,6 +161,37 @@ sealed abstract class TreeList[+A] {
   }
 
   /**
+   * keep the elements that match a predicate
+   */
+  final def filter(fn: A => Boolean): TreeList[A] = {
+    val as = toIterator
+    var resList = List.empty[A]
+    var changed = false
+    while(as.hasNext) {
+      val a = as.next()
+      if (fn(a)) {
+        resList = a :: resList
+      } else { changed = true }
+    }
+    if (changed) TreeList.fromListReverse(resList)
+    else this
+  }
+
+  final def filterNot(fn: A => Boolean): TreeList[A] = {
+    val as = toIterator
+    var resList = List.empty[A]
+    var changed = false
+    while(as.hasNext) {
+      val a = as.next()
+      if (!fn(a)) {
+        resList = a :: resList
+      } else { changed = true }
+    }
+    if (changed) TreeList.fromListReverse(resList)
+    else this
+  }
+
+  /**
    * Standard flatMap on a List type.
    */
   final def flatMap[B](fn: A => TreeList[B]): TreeList[B] = {
@@ -189,6 +230,17 @@ sealed abstract class TreeList[+A] {
     }
     TreeList.fromListReverse(res)
   }
+
+  /**
+   * If the given index is in the list, update and return Some(updated).
+   * else return None
+   *
+   * O(log N)
+   */
+  final def updated[A1 >: A](idx: Long, value: A1): Option[TreeList[A1]] = {
+    val up = updatedOrThis(idx, value)
+    if (up eq this) None else Some(up)
+  }
 }
 
 object TreeList extends TreeListInstances0 {
@@ -199,9 +251,41 @@ object TreeList extends TreeListInstances0 {
     object Nat {
       case class Succ[P <: Nat](prev: P) extends Nat {
         val value: Int = prev.value + 1
+        override def hashCode: Int = value
+        // We don't want to use recursive equality matching
+        // since that would be O(N) for a Nat of value N
+        override def equals(that: Any): Boolean =
+          that match {
+            case s: Succ[_] => value == s.value
+            case _ => false
+          }
+
       }
       case object Zero extends Nat {
         def value: Int = 0
+      }
+
+      private[this] val memoUpTo: Int = 32
+      private[this] val memoNat: Array[Nat] = {
+        @tailrec
+        def build(n: Nat, acc: List[Nat], cnt: Int): Array[Nat] =
+          if (cnt >= memoUpTo) acc.reverse.toArray
+          else {
+            val s = Succ(n)
+            build(s, s :: acc, cnt + 1)
+          }
+
+        build(Zero, Nil, 0)
+      }
+
+      /**
+       * This is a memoized Succ constructor since
+       * we generally only build a small number of Succ instances
+       */
+      def succ[N <: Nat](n: N): Succ[N] = {
+        val v = n.value
+        if (v < memoUpTo) memoNat(v).asInstanceOf[Succ[N]]
+        else Succ(n)
       }
     }
     sealed abstract class NatEq[A <: Nat, B <: Nat] {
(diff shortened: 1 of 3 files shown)


johnynek:
looks like I am raising the coverage by 8.5% with this PR. :)

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -0,0 +1,734 @@
...
+        new NatEq[A, A] {
+          def subst[F[_ <: Nat]](f: F[A]): F[A] = f
+        }
+
+      // Cache this so we avoid allocating repeatedly
+      private[this] val someRefl: Option[NatEq[Nat.Zero.type, Nat.Zero.type]] =
+        Some(refl[Nat.Zero.type])
+
+      def maybeEq[N1 <: Nat, N2 <: Nat](n1: N1, n2: N2): Option[NatEq[N1, N2]] =
+        // I don't see how to prove this in scala, but it is true
```
larsrh: I agree with your comment, but let's have @milessabin take a look to be sure ☺

johnynek:
yeah, this is pretty remarkable:

https://github.com/typelevel/cats-collections/pull/152/files#diff-1f70710e148c78ab48fdd3687d6ecd0bR248

I have often had stack overflows on monads that are not stack safe by default. For instance, in cats we really hurt the performance of State by forcing it inside Eval, and this is done purely for stack safety.

Since the depth is only log N here, you can see we don't blow the stack doing a sequence on Function1 which is the reader monad.

johnynek:
PS: @larsrh can you let me know what I need to do to get this merged? Miles may be busy, so are we blocking merging this on Mile's comment? If so, can I propose instead that we invite Miles to send a follow up PR since this has already been a lot of work (which of course I signed up for, but I do hope it will come to some productive end).

Thanks.

larsrh:
@johnynek No need to block on a response from @milessabin.

Because it's a substantial contribution, I'd like two wait for two things:
* [ ] @LukaJCB self-requested a review, so he's probably preparing to leave some comments soon
* [ ] I'd also like to give it a reviewing pass, which I'll hopefully complete by Sunday.

jozic:
Would be nice to have some benchmarks to compare with Scala's List, as well as the example where Scala's List does blow the stack

Also would be nice to get a doc page, similar to https://github.com/typelevel/cats-collections/blob/master/docs/src/main/tut/range.md

larsrh:
@jozic Both things are nice to have, but could also go into follow-up PRs.

jozic:
@larsrh 
that's exactly what I meant by saying `nice to have` :)

johnynek:
@jozic I added a test that shows List blowing the stack and this not. I linked to it in this comment:

https://github.com/typelevel/cats-collections/pull/152#issuecomment-450700104

I think we should follow up with benchmarks. This is already a very large PR, larger than I would like, but sending partial functionality in this case seemed worse, but I could split it into a more evolutionary series of PRs, though I think I’d have to make the private Impl public for testing in that case.

larsrh:
I think it's fine to keep it this way. More tests/benchmarks can go into follow-up PRs.

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -0,0 +1,867 @@
...
+  def toList: List[A]
+
+  /**
+   * Convert to a scala standard list, but reversed
+   */
+  def toListReverse: List[A]
+
+  /**
+   * return the right most full binary tree on the right, the rest on left
+   * val (l, r) = items.split
```
LukaJCB: AFAICT this isn't being run right now, correct? We could easily turn this into a doctest though, WDYT? :)

johnynek: I don't use doctests, so I don't know how to set this up. I do have a test of this law:

https://github.com/typelevel/cats-collections/pull/152/files#diff-1f70710e148c78ab48fdd3687d6ecd0bR91

What do you think about someone who is experienced with doctest sending that as a follow up? Perhaps they can add doctests for many of the methods.

I can remove the comment, if you prefer not to have the code there without a doctest.

johnynek:
@LukaJCB thank you for taking the time to review my code.

core/src/main/scala/cats/collections/TreeList.scala
LukaJCB: I'll do a follow up :)

tests/src/test/scala/cats/collections/TreeListSpec.scala
```diff
@@ -0,0 +1,273 @@
...
+    (0L until xs.size).foreach { idx =>
+      assert(xs.get(idx) == Some(list(idx.toInt)))
+    }
+  })
+
+  test("toIterator throws the same type of exception as List on empty")(forAll { (xs: TreeList[Int]) =>
+    val it = xs.toIterator
+    // exhaust the iterator
+    it.size
+    assertThrows[NoSuchElementException] { Nil.iterator.next }
```
jozic: this can be outside of forAll, right?

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -0,0 +1,867 @@
...
+      res = revTrees.next() :: res
+    }
+    res
+  }
+
+  /**
+   * Take the first n things off the list. O(n)
+   */
+  final def take(n: Long): TreeList[A] = {
+    val takeIt = toIterator
```
jozic: if n > size we could just return this, 
perhaps a micro-optimization

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -0,0 +1,867 @@
...
+    }
+  }
+
+  import Impl._
+
+  private final case class Trees[A](treeList: List[Tree[Nat, A]]) extends TreeList[A] {
+    def prepend[A1 >: A](a1: A1): TreeList[A1] =
+      treeList match {
+        case h1 :: h2 :: rest =>
+          def go[N1 <: Nat, N2 <: Nat, A2 <: A](t1: Tree[N1, A2], t2: Tree[N2, A2]): TreeList[A1] =
```
jozic: just curious, what's the purpose of having `go`? it's not recursive and is invoked only once right after definition

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -0,0 +1,867 @@
...
+        treea.foldRight(b)(fn)
+      }
+
+    def isEmpty: Boolean = treeList.isEmpty
+    def nonEmpty: Boolean = treeList.nonEmpty
+
+    def map[B](fn: A => B): TreeList[B] = Trees(treeList.map(_.map(fn)))
+
+    def drop(n: Long): TreeList[A] = {
+      @tailrec
```
jozic: another micro-optimization is to check if n > size and return empty
benchmarks probably can reveal if it's worth it or not

core/src/main/scala/cats/collections/TreeList.scala
johnynek: since computing size itself is `O(log N)`, I'm going to leave this off for now until we are more sure we have evidence it is faster. If people usually take small numbers of things, it could be a deoptimization on large lists.

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -0,0 +1,867 @@
...
+  private object Impl {
+    sealed trait Nat {
+      def value: Int
+    }
+    object Nat {
+      case class Succ[P <: Nat](prev: P) extends Nat {
+        val value: Int = prev.value + 1
+      }
+      case object Zero extends Nat {
+        def value: Int = 0
```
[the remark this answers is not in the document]
johnynek: since it is returning a constant, it does not need to be. in fact, I think it only makes the object take more space if it does.

The Succ case needs a val otherwise it is O(value) cost to compute value.

core/src/main/scala/cats/collections/TreeList.scala
johnynek: I wanted to be able to name the types for better clarity: `N1` and `N2`. That's the only reason. I'll comment.

core/src/main/scala/cats/collections/TreeList.scala
jozic: so my understanding is that List will have at most 3 tree elements and for Root size is constant call (1) and for Balanced size is cached in already calculated field
so calculating TreeLIst size will take 3 iterations (at most) and adding 3 ints which looks O(1) to me

i guess i'm missing something, sorry for taking your time on that

jozic: Like having a TreeLIst of 101 elements
```val tl = TreeList.fromList((0 to 100).toList)```

I see this

image

File: core/src/main/scala/cats/collections/TreeList.scala [hunk 8]
```diff
@@ -0,0 +1,889 @@
...
+  /**
+   * return the right most full binary tree on the right, the rest on left
+   * val (l, r) = items.split
+   * assert((l ++ r) == items)
+   *
+   * O(log N)
+   */
+  def split: (TreeList[A], TreeList[A])
+
+  /*
```
jozic: minor, 
misses second * to be considered a proper JavaDoc/ScalaDoc

core/src/main/scala/cats/collections/TreeList.scala
johnynek: I don't think it is true that we have O(1) items in the list. I think we have O(log N) items. Since each tree is fully balanced, the size is `2^n - 1` for some n. So, I think for various sizes you have:
```
0 -> []
1 -> [1]
2 -> [1, 1]
3 -> [3]
4 -> [1, 3]
5 -> [1, 1, 3]
6 -> [3, 3]
7 -> [7]
8 -> [1, 7]
...
12 -> [1, 1, 3, 7]
...
```
Note, 12 has 4 items in the list.

tests/src/test/scala/cats/collections/TreeListSpec.scala
jozic: if it's fast, then should be good as is

jozic:
@johnynek 
my only concern is I'm missing how size is O(log N) :)  otherwise looks very neat
thanks for taking time to address my nitpicks

core/src/main/scala/cats/collections/TreeList.scala
jozic: I think i've got it myself
it's not 3, it's indeed long N

johnynek:
Thank you @larsh

larsrh:
Thanks, Oscar!

PR #160, opened 2019-01-26
improve Heap (fix #158), increase test coverage

### LINKED ISSUE #158: Heap.heapify appears to be O(N^2)
https://github.com/typelevel/cats-collections/blob/master/core/src/main/scala/cats/collections/Heap.scala#L71

This method is claimed (without proof) to be O(N). But it is calling in a loop `.length` and accessing elements from a list, both of which are O(N) operations.

In the very best case I think this can be O(N log N), but I think in fact it is O(N^2).

I think converting the input List into a mutable java `Array[A]` would fix the performance issue since we don't need to update that list.

Incidentally, you could also widen the input type to `Iterable[A]` I think since we should copy into an O(1) indexable structure in the method.

That would regain the O(N) complexity I believe (or again, maybe O(N log N) in the worst case).

cc @anicolaspp

johnynek:
This does three things:

1. increase test coverage (I want to go for 100% so I will probably add a few more commits after codecov runs).
2. fix #158 by using an Array rather than indexing into a List
3. notice that heapify totally ignores the current `Heap` so it should be on the companion. If we can break compatibility, we should remove the method.

I'm not super excited about the Heap code, if we can break compatibility, I'd like to make a few other changes. Please advise on that and I will follow up in a second PR.

### DIFF
--- a/core/src/main/scala/cats/collections/Heap.scala
+++ b/core/src/main/scala/cats/collections/Heap.scala
@@ -64,21 +64,12 @@ sealed abstract class Heap[A] {
     else
       bubbleUp(min, left.add(x), right)
 
-  /**
-   * Build a heap using a list.
-   * Order O(n)
-   */
-  def heapify(a: List[A])(implicit order: Order[A]): Heap[A] = {
-    def loop(i: Int, xs: scala.List[A]): Heap[A] =
-      if (i < xs.length) {
-        bubbleDown(xs(i), loop(2 * i + 1, xs), loop(2 * i + 2, xs))
-      }
-      else {
-        Leaf()
-      }
 
-    loop(0, a)
-  }
+  /**
+   * Avoid this, it should really have been on the companion
+   */
+  def heapify(a: List[A])(implicit order: Order[A]): Heap[A] =
+    Heap.heapify(a)
 
   /**
    * Remove the min element from the heap (the root).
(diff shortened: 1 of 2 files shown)


johnynek:
@larsrh what do you think of this?

core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -55,30 +55,21 @@ sealed abstract class Heap[A] {
...
-      if (i < xs.length) {
-        bubbleDown(xs(i), loop(2 * i + 1, xs), loop(2 * i + 2, xs))
-      }
-      else {
-        Leaf()
-      }
-
-    loop(0, a)
-  }
+  def heapify(a: List[A])(implicit order: Order[A]): Heap[A] =
```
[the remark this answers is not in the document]
larsrh: Let's have a deprecation cycle there.

tests/src/test/scala/cats/collections/HeapSpec.scala
```diff
@@ -1,16 +1,19 @@
...
+import org.scalacheck.{Arbitrary, Gen}
 
 /**
  * Created by nperez on 3/28/16.
  */
 class HeapSpec extends CatsSuite {
-  test("sorted")(
-    forAll { (xs: scala.List[Int]) =>
+  implicit val propConfig =
+    PropertyCheckConfig(minSuccessful = 200)
```
larsrh: I recommend dialing that up. Some errors in `Diet` (#150) were only found after thousands of iterations.

```suggestion
    PropertyCheckConfig(minSuccessful = if (Platform.isJvm) PosInt(10000) else PosInt(500))
```

tests/src/test/scala/cats/collections/HeapSpec.scala
```diff
@@ -19,4 +22,131 @@ class HeapSpec extends CatsSuite {
...
+        } yield heap + a
+      }
+      else Gen.const(Heap.empty[A])
+
+    Gen.frequency((2, addOnly), (3, startWith1), (5, heapify), (1, addMoreAndRemove), (1, smallerAdd))
+  }
+
+  implicit def arbHeap[A: Arbitrary: Order]: Arbitrary[Heap[A]] =
+    Arbitrary {
+      Gen.choose(0, 10000).flatMap(heapGen[A](_, Arbitrary.arbitrary[A]))
```
larsrh: This should probably use the `sized` combinator. If we want large test sizes, then we can also increase that in the config.

tests/src/test/scala/cats/collections/HeapSpec.scala
johnynek: It takes about 10s I think at 200, at 10000 this would be about 5 minutes of testing.

Are you sure you want that long?

I think the problem is the Gen is pretty expensive to try to simulate Heaps at all possible states. I think that needs to be improved if we should turn it up.

How much of a time budget would you allow for this test?

larsrh: 5 minutes would indeed be excessive. Let's say 500 on the JVM and 100 on JS, then.

johnynek:
@larsrh okay, I think this is ready to merge. Let me know what you think.

I'll follow up with the implementation of `UnorderedFoldable[Heap]` and some related methods on `Heap` (e.g. exists, forall, unorderedFoldMap, etc...)

larsrh:
Thanks a lot for putting in the effort!

PR #171, opened 2019-01-29
tighter Diet invariant

larsrh:
Follow-up to #150


larsrh:
The tighter invariant actually does not hold, as evidenced by this example:

```scala
scala> Diet.empty[Int] + Range(1, 3) - 2 + 2
res5: cats.collections.Diet[Int] = DietNode(Range(1,2),EmptyDiet,DietNode(Range(3,3),EmptyDiet,EmptyDiet))
```

Pretty sure this broke in 93ccd74f8774cde77e2b33bdcac5bc2b806dcb50.

PR #173, opened 2019-01-31
Add Heap#addAll and stronger invariant test

johnynek:
I added a stronger test of the fully balanced tree invariant in Heap.

I added addAll and a synonym ++

finally, I tried to implement merge, but I couldn't find anything that kept the strong invariants and was any better than O(N log N) on the smaller heap, which is trivial to implement but I don't think useful since it is the same as `left ++ (right.toList)` in complexity.

I think it may be impossible to make merge efficient while also maintaining a completely balanced tree without any slack. It might be possible to show worst case is O(N log N) by constructing some pathological examples of pairs of highly unbalanced heaps.

cc @larsrh

### DIFF
--- a/core/src/main/scala/cats/collections/Heap.scala
+++ b/core/src/main/scala/cats/collections/Heap.scala
@@ -70,6 +70,18 @@ sealed abstract class Heap[A] {
     }
 
   /**
+   * Add a collection of items in. This is O(N log N) if as is size N
+   */
+  def addAll(as: Iterable[A])(implicit order: Order[A]): Heap[A] = {
+    val ait = as.iterator
+    var heap = this
+    while(ait.hasNext) {
+      heap = heap + ait.next()
+    }
+    heap
+  }
+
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -69,6 +69,18 @@ sealed abstract class Heap[A] {
         bubbleUp(min, left.add(x), right)
     }
 
+  /**
+   * Add a collection of items in. This is O(N log N) if as is size N
+   */
+  def addAll(as: Iterable[A])(implicit order: Order[A]): Heap[A] = {
+    val ait = as.iterator
+    var heap = this
+    while(ait.hasNext) {
```
vkostyukov: nit: space is missing after `while`.

johnynek:
@larsrh given the original author here @vkostyukov likes this PR.

Any concerns about merging it?

PR #174, opened 2019-01-31
Add PairingHeap for a fast Heap CommutativeMonoid

### LINKED ISSUE #165: efficient merges for Heaps
we can always merge two heaps by taking the smaller of the two and adding each element. This would be `n log n` where n is the size of the smaller heap. This totally discards the structure of the smaller one.

It *seems* like we should be able to leverage the heap property to more efficiently merge (say O(log n) would be ideal).

The original paper this is based on does not have a merge operation, but there may be other heap algorithms or small changes we can make to this implementation to support merge.

johnynek:
closes #165 

Also adds a utility method `takeLargest` which is big use of heaps: taking the top-k elements.

cc @ttim @larsrh

### DIFF
--- a/core/src/main/scala/cats/collections/Heap.scala
+++ b/core/src/main/scala/cats/collections/Heap.scala
@@ -172,6 +172,17 @@ sealed abstract class Heap[A] {
    */
   def --(implicit order: Order[A]): Heap[A] = remove
 
+  /**
+   * convert to a PairingHeap which can do fast merges,
+   * this is an O(N) operation
+   */
+  def toPairingHeap: PairingHeap[A] =
+    if (isEmpty) PairingHeap.empty
+    else {
+      val thisBranch = this.asInstanceOf[Branch[A]]
+      import thisBranch.{min, left, right}
+      PairingHeap.Tree(min, left.toPairingHeap :: right.toPairingHeap :: Nil)
+    }
 }
 
 object Heap {
--- a/core/src/main/scala/cats/collections/Heap.scala
+++ b/core/src/main/scala/cats/collections/Heap.scala
@@ -21,9 +21,14 @@ sealed abstract class Heap[A] {
   import Heap._
 
   /**
-   * Returns min value on the heap.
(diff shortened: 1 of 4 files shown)


johnynek:
I'll follow up this PR with a benchmarks PR, but here is a preview:
```
[info] Benchmark                         (n)   Mode  Cnt      Score       Error  Units
[info] HeapBench.addHeap                 100  thrpt    3  37691.149 ± 35811.213  ops/s
[info] HeapBench.addHeap                1000  thrpt    3   2335.278 ±   386.613  ops/s
[info] HeapBench.addHeap               10000  thrpt    3    172.421 ±    12.908  ops/s
[info] HeapBench.addPairingHeap          100  thrpt    3  82610.811 ±  6063.224  ops/s
[info] HeapBench.addPairingHeap         1000  thrpt    3    548.521 ±    12.160  ops/s
[info] HeapBench.addPairingHeap        10000  thrpt    3      8.056 ±     1.788  ops/s
[info] HeapBench.removeAllHeap           100  thrpt    3  25568.689 ±  7525.437  ops/s
[info] HeapBench.removeAllHeap          1000  thrpt    3   1301.152 ±   115.899  ops/s
[info] HeapBench.removeAllHeap         10000  thrpt    3     83.451 ±    31.589  ops/s
[info] HeapBench.removeAllPairingHeap    100  thrpt    3  43445.061 ±  7571.888  ops/s
[info] HeapBench.removeAllPairingHeap   1000  thrpt    3    877.610 ±  1331.924  ops/s
[info] HeapBench.removeAllPairingHeap  10000  thrpt    3     14.978 ±     6.040  ops/s
```

It looks like PairingHeap is faster than our current Heap at sizes around 500 or so (I'll rerun with more iterations and more sizes to find the cut-off), but after that point Heap is considerably faster.

It could be a bug with my implementation of course, that only hurts performance, not correctness (since we have very good test coverage, I believe the main properties we want are true, but I could have made a mistake in implementing combineAll, which perhaps results in correct, but slower heaps).

PR #177, opened 2019-02-01
Add a fast immutable BitSet implementation

johnynek:
This is an implementation of a fast immutable BitSet (based on a Tree and `Array[Long]`) that @non and I did about a year back.

Here are some benchmarks:
```
[info] Benchmark                      (density)  (exponent)  Mode  Cnt   Score    Error  Units
[info] BitSetBenchmarks.buildCcbs           0.5          10  avgt    3   4.577 ±  4.389  us/op
[info] BitSetBenchmarks.buildIset           0.5          10  avgt    3  81.805 ± 30.429  us/op
[info] BitSetBenchmarks.buildSci            0.5          10  avgt    3  65.580 ± 38.807  us/op
[info] BitSetBenchmarks.buildScm            0.5          10  avgt    3  90.134 ± 20.532  us/op
[info] BitSetBenchmarks.foldIntoCcbs        0.5          10  avgt    3  25.780 ±  2.740  us/op
[info] BitSetBenchmarks.foldIntoIset        0.5          10  avgt    3  84.137 ±  5.968  us/op
[info] BitSetBenchmarks.foldIntoSci         0.5          10  avgt    3  42.205 ±  4.332  us/op
[info] BitSetBenchmarks.foldIntoScm         0.5          10  avgt    3   8.769 ±  0.066  us/op
[info] BitSetBenchmarks.lookupCcbs          0.5          10  avgt    3   0.703 ±  0.112  us/op
[info] BitSetBenchmarks.lookupIset          0.5          10  avgt    3   1.206 ±  0.082  us/op
[info] BitSetBenchmarks.lookupSci           0.5          10  avgt    3   0.839 ±  0.140  us/op
[info] BitSetBenchmarks.lookupScm           0.5          10  avgt    3   0.824 ±  0.173  us/op
[info] BitSetBenchmarks.mergeCcbs           0.5          10  avgt    3   0.045 ±  0.006  us/op
[info] BitSetBenchmarks.mergeIset           0.5          10  avgt    3  14.797 ±  2.277  us/op
[info] BitSetBenchmarks.mergeSci            0.5          10  avgt    3   0.141 ±  0.252  us/op
[info] BitSetBenchmarks.mergeScm            0.5          10  avgt    3   0.057 ±  0.007  us/op
```

where you can see a comparison to `immutable.BitSet  = Sci`, `mutable.BitSet = Scm` and `immutable.Set[Int] = Iset`. As you can see, the performance the best except for fold, where mutable wins. But otherwise *this implementation beats all the rest*.

This code needs comments, and we should add some instances, e.g. `CommutativeMonoid[BitSet]`, `Semiring[BitSet]`, `GenBool[BitSet]`.

Do you think this is suitable for inclusion with some improvements @larsrh ?

### DIFF
--- a/bench/src/main/scala/BitSetBenchmarks.scala
+++ b/bench/src/main/scala/BitSetBenchmarks.scala
@@ -1,0 +1,131 @@
+package cats.collections
+
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+
+import scala.collection.{immutable, mutable}
+import scala.math.pow
+import scala.util.Random
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+class BitSetBenchmarks {
+
+  // ranges from 0.0 to 1.0.
+  @Param(Array("0.5"))
+  var density: Double = _
+
+  // creates bitsets of size 2^k (2^10 = 1024, 2^20 = ~1M).
+  @Param(Array("10"))
+  var exponent: Int = _
+
+  var size: Int = _
+  var indices: Array[Int] = _
+
+  var values: Array[Int] = _
+  var sci: immutable.BitSet = _
+  var scm: mutable.BitSet = _
+  var iset: immutable.Set[Int] = _
+  var ccbs: _root_.cats.collections.BitSet = _
+
+  var values2: Array[Int] = _
+  var sci2: immutable.BitSet = _
+  var scm2: mutable.BitSet = _
+  var iset2: immutable.Set[Int] = _
+  var ccbs2: _root_.cats.collections.BitSet = _
+
+  @Setup
+  def setup(): Unit = {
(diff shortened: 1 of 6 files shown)


johnynek:
looks like some of the performance geek-outs fail the linter. I think they are worth it given how hard we worked to get the performance to be superior to everything else, yet still immutable.

larsrh:
I say we kill the linter (or reconfigure).

non:
@johnynek This looks good to me. 👍 

I'm gonna go write a bunch of comments in a commit that you can pull in (unless you're already doing it).

non:
I added a bunch of comments in https://github.com/non/cats-collections/commit/76aff97cb9872538a5055466a69a00ad22e986e8.

johnynek:
@larsrh what do you think? @non and I would like to merge and iterate a few more times (add some more benchmarks, possibly add a few optimizations we've found offline, e.g. Int looks like it is a better internal representation that Long is, and will probably be much faster on scalajs).

PR #178, opened 2019-02-01
Two commits for oscar/bitset

non:
The first commit adds a bunch of comments.

The second expands the range we're testing over (we weren't testing negative integers even though they were totally supported).

### DIFF
--- a/core/src/main/scala/cats/collections/BitSet.scala
+++ b/core/src/main/scala/cats/collections/BitSet.scala
@@ -6,29 +6,159 @@ import scala.annotation.tailrec
 import BitSet.{Branch, Empty, Leaf}
 
 /**
- * A fast immutable BitSet which unlike scala's default immutable BitSet does
- * not do a full copy on each add. Interally it is based on a Tree of Array[Long].
- * The benchmarks suggest this is MUCH faster for cases where you need many modifications
- * and merges, e.g. a BloomFilter.
+ * A fast, immutable BitSet.
+ *
+ * A Bitset is a specialized type of set that tracks the `Int` values
+ * it contains: for each integer value, a BitSet use a single bit to
+ * track whether it is present (1) or absent (0). An empty bitset is a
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/collections/BitSet.scala
```diff
@@ -75,6 +235,9 @@ sealed abstract class BitSet { lhs =>
       override def empty = Empty.toSet
     }
 
+  /**
+   * Returns false this bitset contains values, true otherwise.
```
johnynek: False *if* this bitset... ?

core/src/main/scala/cats/collections/BitSet.scala
```diff
@@ -86,11 +249,26 @@ sealed abstract class BitSet { lhs =>
         }
     }
 
+  /**
+   * Returns true this bitset contains values, false otherwise.
```
johnynek: True *if* this ...

core/src/main/scala/cats/collections/BitSet.scala
```diff
@@ -117,6 +306,13 @@ sealed abstract class BitSet { lhs =>
...
     }
 
+  /**
+   * Universal hash code.
+   *
+   * Bitsets that are the equal will hash to the same value. As in
+   * `equals`, the values present determine the hash code, as opposed
+   * to the tree structure.
+   */
   override def hashCode: Int = {
```
johnynek: Maybe this should be a lazy Val...

non: I'm not 100% convinced, so let's wait on that.

core/src/main/scala/cats/collections/BitSet.scala
non: Fixed!

core/src/main/scala/cats/collections/BitSet.scala
non: Fixed!

PR #180, opened 2019-02-02
Add xor and set difference operators to BitSet

non:
These are operations that are a bit tricky to implement but would likely be very useful to many users.

I also went ahead and added `Order` and `Ordering` instances, although I held off on other type class instances for now.

### DIFF
--- a/tests/src/test/scala/cats/collections/BitSetTest.scala
+++ b/tests/src/test/scala/cats/collections/BitSetTest.scala
@@ -1,7 +1,7 @@
 package cats.collections
 
 import org.scalacheck.Prop._
-import org.scalacheck.{Arbitrary, /*Cogen, */Gen, Properties, Test}
+import org.scalacheck.{Arbitrary, Properties, Test}
 import Arbitrary.{arbitrary => arb}
 
 object BitSetTest extends Properties("BitSet") {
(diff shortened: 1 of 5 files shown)


tests/src/test/scala/cats/collections/BitSetTest.scala
```diff
@@ -176,4 +177,96 @@ object BitSetTest extends Properties("BitSet") {
...
+      val rhs = (y ^ x)
+      (lhs == rhs) :| s"$lhs == $rhs"
+    }
+
+  property("(x ^ x) = 0") =
+    forAll { (x: BitSet) =>
+      val lhs = (x ^ x)
+      val rhs = BitSet.empty
+      (lhs == rhs) :| s"$lhs == $rhs"
+    }
```
johnynek: can we add a law like:
```scala
(x ^ y)(z) == (x(z) ^ y(z))
```

also, should we just use the law testing in cats-algebra vs re-inlining? See https://github.com/typelevel/cats-collections/pull/172 -- it's pretty easy, but I think we need a new test class to use the laws.

johnynek:
I squash merged your previous PR, can you rebase, sorry?

tests/src/test/scala/cats/collections/BitSetTest.scala
non: I support using the law-testing, but I'd rather do that in a follow-up if that's OK. I'll add this law.

PR #182, opened 2019-02-03
Add a PartiallyOrderedSet typeclass

### LINKED ISSUE #176: should we have a Heap typeclass?
I think we can have a `IsHeap[F[_]]` typeclass. There are many heap implementations, and we can convert the current tests into a HeapLaws to test different implementations.

something like:
```scala
trait IsHeap[F[_]] extends UnorderedFoldable[F] {
  def add[A](fa: F[A], a: A)(implicit ord: Order[A]): F[A]
  def remove[A](fa: F[A])(implicit ord: Order[A]): F[A]
  // this could be a method on UnorderedFoldable, but it was incorrectly placed on Foldable in cats
  def minimumOption[A](fa: F[A]): Option[A]
  def toList[A](fa: F[A])(implicit ord: Order[A]): List[A]
  def foldLeft[A, B](fa: F[A], init: B)(fn: (B, A) => B)(implicit ord: Order[A]): B
}
```

johnynek:
closes #176 

What do you think @larsrh @vkostyukov ?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ lazy val buildSettings = Seq(
 lazy val `cats-collections` = project.in(file("."))
   .settings(buildSettings:_*)
   .settings(noPublishSettings)
-  .aggregate(coreJVM, coreJS, bench, scalacheckJVM, scalacheckJS, testsJVM, testsJS, docs)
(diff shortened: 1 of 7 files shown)


larsrh:
This conflicts post-#173 and needs a rebase.

johnynek:
merged with master and added implementations for PairingHeap.

PR #186, opened 2019-02-25
Add Show instance and fromFoldable to AvlSet

ybasket:
This PR adds two small convenience features to `AvlSet`:
1. An instance of `Show`, implementation based on the one `cats-core` defines for Scala's `Set`
2. A `fromFoldable` on the companion object allowing to construct `AvlSet`s from more collections than just `List`, for example `Chain`.

### DIFF
--- a/core/src/main/scala/cats/collections/Set.scala
+++ b/core/src/main/scala/cats/collections/Set.scala
@@ -295,7 +295,7 @@ sealed abstract class AvlSet[A] {
   }
 
   override def toString: String =
-    "Set(" + Foldable[List].intercalate(toList.map(_.toString), ",") + ")"
+    "AvlSet(" + Foldable[List].intercalate(toList.map(_.toString), ", ") + ")"
 
   // So yeah. we had to make a decision, either we have to make this
   // structure Key/Value pairs even when we don't always need a value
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/collections/Set.scala
```diff
@@ -418,4 +424,9 @@ trait AvlSetInstances {
     override def eqv(x: AvlSet[A], y: AvlSet[A]): Boolean =
       iteratorEq(x.toIterator, y.toIterator)
   }
+
+  implicit def showSet[A: Show]: Show[AvlSet[A]] = new Show[AvlSet[A]] {
+    override def show(t: AvlSet[A]): String =
+      t.toIterator.map(_.show).mkString("AvlSet(", ", ", ")")
```
LukaJCB: Can you add a quick property test for this one, that compares this with `toString` for an `A` where `show` and `toString` are the same? (i.e. `Int` or similar)

ybasket: Sure, just added. For it to succeed, I had to change the output of the `Show` instance to match the `toString`, which outputs something like _Set(x,y,z)_ instead of _AvlSet(x, y, z)_ which I had before (and which would be consistent with cats-core). Any preference/constraint on what formatting to use?

larsrh: Sorry for the back-and-forth: I think what @LukaJCB meant was to change `toString` such that it _also_ produces `AvlSet(x,y,z)`.

ybasket: Luckily not a difficult commit - changed it to match the cats-core formatting.

larsrh:
Thanks for your contribution!

LukaJCB:
Yes, thanks! :)

PR #195, opened 2019-05-31
Update sbt and plugins

jhnsmth:


larsrh:
This probably requires an update to 2.13.0-RC2, too.

jhnsmth:
@larsrh This was my initial motivation, but I  skipped it as algebra is not published for RC2.
But the problem with the build is different, starting from v8.0 `sbt-microsites` added support for `mdoc` which is not published for scala 2.13 at all,  see 47deg/sbt-microsites#305. So the options here are  
- disable docs building for 2.13 
- downgrade the plugin back to latest release of 0.7.x which is 0.7.27
I think I'm in favor of the second one.

larsrh:
I have no strong opinion, so feel free to follow along your preferred choice.

jhnsmth:
@larsrh thanks, I've updated PR

larsrh:
Thanks!

Issue #200, opened 2019-06-15
Rename ISet to Predicate?


xplosunn:
`Predicate` seems to be a more common name for the same concept. Would the rename make sense here?

larsrh:
I'm personally in favour, but I have no strong opinions.

PR #201, opened 2019-06-16
Rename ISet to Predicate

xplosunn:
Solves #200


core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -0,0 +1,75 @@
...
+object Predicate extends PredicateInstances {
+  def apply[A](f: A => Boolean): Predicate[A] = new Predicate[A] {
+    def apply(a: A) = f(a)
+  }
+
+  def empty: Predicate[Any] = apply(_ => false)
+}
+
+trait PredicateInstances {
+  implicit def predicateMonoid[A]: Monoid[Predicate[A]] = new Monoid[Predicate[A]] {
```
johnynek: It’s hard for me to see there should be an implicit monoid. Union and intersection are reasonable choices. I think users should set them up or use class wrappers/opaque type to choose.

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Returns the predicate which is the the difference of another predicate removed from this predicate
+   */
+  def diff[B <: A](remove: Predicate[B]): Predicate[B] = Predicate(a => apply(a) && !remove(a))
+
+  /**
+   * Returns the predicate which is the the difference of another predicate removed from this predicate
+   */
+  def -[B <: A](remove: Predicate[B]): Predicate[B] = Predicate(a => apply(a) && !remove(a))
```
johnynek: Can we just call diff?

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * returns a predicate which is the intersection of this predicate and another
+   */
+  def intersection[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) && other(a))
+
+  /**
+   * returns a predicate which is the intersection of this predicate and another
+   */
+  def &[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) && other(a))
```
johnynek: Can we just call intersection?

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -0,0 +1,75 @@
...
+abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
+  /**
+   * returns a predicate which is the union of this predicate and another
+   */
+  def union[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) || other(a))
+
+  /**
+   * returns a predicate which is the union of this predicate and another
+   */
+  def |[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) || other(a))
```
johnynek: Can we call union?

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Return the opposite predicate
+   */
+  def negate: Predicate[A] = Predicate(a => !apply(a))
+
+  /**
+   * Return the opposite predicate
+   */
+  def unary_!(): Predicate[A] = negate
```
johnynek: Would be nice to have a contraMap here and a Contravariant instance.

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -0,0 +1,75 @@
...
+trait PredicateInstances {
+  implicit def predicateMonoid[A]: Monoid[Predicate[A]] = new Monoid[Predicate[A]] {
+    override def empty: Predicate[A] = Predicate.empty
+    override def combine(l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
+  }
+
+  implicit val predicateInstance: MonoidK[Predicate] = new MonoidK[Predicate] {
+    override def empty[A]: Predicate[A] = Predicate.empty
+    override def combineK[A](l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
+  }
```
johnynek: I think we can also make Semigroupal instances which do union or intersection.

xplosunn:
I just renamed and kept all the methods / instances. If @larsrh or someone else agrees with the changes I don't mind doing them.

larsrh:
@johnynek All of those things are problems with the current `ISet`; they haven't been introduced in this PR. As such, I'm going to merge this and open a new issue to work on these things.

larsrh:
Thanks @xplosunn!

PR #207, opened 2019-06-27
Scala 2.13

larsrh:
~Doesn't compile at the moment because the catalysts dependency is unavailable.~


larsrh:
I will update this once a new release of sbt-crossproject is out with portable-scala/sbt-crossproject#108 merged.

Issue #210, opened 2019-06-28
Make data structures covariant?


ceedubs:
Some (most?) of the structures in cats-collections, such as `Diet` and `Range` are covariant in nature, but not by definition. This leads to some minor awkwardness like [asInstanceOf calls](https://github.com/typelevel/cats-collections/blob/8089582ed7fc61f9c23a0bf76ac3401585e1a98e/core/src/main/scala/cats/collections/Diet.scala#L287). It also prevents 3rd party data types that contain these structures from being declared as covariant, which there isn't really a good workaround for.

I've gone back and forth on variance in Scala, but over time I've settled on the opinion that variance is useful in Scala, it isn't really problematic as long as you avoid things like universal equality, and that you shouldn't fight it. If cats-collections maintainers want to stick to invariance, that's okay but I wanted to ask about this and would be happy to do legwork in conversion if there's interest.

**Edit**: I guess that there's also perhaps a consistency argument since data structures such as `NonEmptyList` in Cats are already covariant.

johnynek:
I personally prefer them to be covariant where possible as well.

larsrh:
Good point. I'd be very happy to accept such a PR.

ceedubs:
I'm going to go ahead and close this out. `Dequeue` is the only data structure currently in cats-collections that I'm convinced is improved by being made covariant, and that was done in #214. These sort of general issues can hang around forever. People can create separate issues for any individual structures that they think should be converted.

PR #211, opened 2019-06-29
[do not merge] Make data structures covariant

ceedubs:
This is a start to #210. I wanted to get a review of this change before going through the work of converting everything that could be converted.

I believe that this should be binary compatible, but it is _not_ source compatible.


larsrh:
> I believe that this should be binary compatible, but it is _not_ source compatible.

Source breakage is minor, so I'm fine with proceeding like this.

ceedubs:
@larsrh would you prefer that I change as many structures as I can in this PR or submit separate PRs?

larsrh:
> would you prefer that I change as many structures as I can in this PR or submit separate PRs?

Feel free to bundle as many as you've got time for ☺

core/src/main/scala/cats/collections/Range.scala
```diff
@@ -125,8 +125,8 @@ final case class Range[A](val start: A, val end: A) {
     * Folds over the elements of the range from right to left; accumulates a value of type B
     * by applying the function f to the current value and the next element.
     */
-  def foldRight[B](s: B, f: (A, B) => B)(implicit discrete: Discrete[A], order: Order[A]): B =
-    reverse.foldLeft(s, (b: B, a: A) => f(a, b))(discrete.inverse, Order.reverse(order))
+  def foldRight[AA >: A, B](s: B, f: (AA, B) => B)(implicit discrete: Discrete[AA], order: Order[AA]): B =
```
johnynek: Really weird choice that f is not a separate parameter group here. Seems not idiomatic.

core/src/main/scala/cats/collections/Map.scala
```diff
@@ -7,7 +7,7 @@ import scala.collection.mutable.ListBuffer
 /**
  * A tree based immutable map.
  */
-class AvlMap[K,V](val set: AvlSet[(K,V)]) {
+class AvlMap[K,+V](val set: AvlSet[(K,V)]) {
```
ceedubs: I only made this covariant in V. I think that we _could_ make it covariant on `K`, but I don't know if it is useful.

larsrh: Probably not.

core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -180,9 +180,10 @@ sealed abstract class Heap[A] {
    * prefer unorderedFoldMap if you can express your operation as a commutative monoid
    * since it is O(N) vs O(N log N) for this method
    */
-  def foldLeft[B](init: B)(fn: (B, A) => B)(implicit order: Order[A]): B = {
+  // TODO ceedubs should this be (B, A) or (B, AA)?
```
johnynek: Since Function2 is contravariant AA is the more general type, since you could also accept A for that case.

core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -194,28 +195,28 @@ sealed abstract class Heap[A] {
...
    */
-  def --(implicit order: Order[A]): Heap[A] = remove
+  def --[AA >: A](implicit order: Order[AA]): Heap[AA] = remove[AA]
 
   /**
    * convert to a PairingHeap which can do fast merges,
    * this is an O(N) operation
    */
-  def toPairingHeap: PairingHeap[A] =
+  def toPairingHeap[AA >: A]: PairingHeap[AA] =
```
johnynek: I don’t see why this needs to be AA if PairingHeap is covariant.

core/src/main/scala/cats/collections/Set.scala
```diff
@@ -14,7 +14,7 @@ import cats.collections.compat.Factory
  * This data-structure maintains balance using the
  * [AVL](https://en.wikipedia.org/wiki/AVL_tree) algorithm.
  */
-sealed abstract class AvlSet[A] {
+sealed abstract class AvlSet[+A] {
```
johnynek: This case is a bit less clear to me. Already this class doesn’t make sense if there are multiple Orderings at different parts of the program. Here we could be magnifying this chance by accepting Orderings on supertypes.

So we now have a stronger requirement: only one Ordering for each type and all super-Orderings giving consistent comparison results. This seems like a nice property but easily overlooked I fear and possibly a source of subtle bugs.

core/src/main/scala/cats/collections/Heap.scala
ceedubs: Oops I meant to grep for TODO before I pushed this. I'll clean this up.

core/src/main/scala/cats/collections/Heap.scala
ceedubs: Hmm yeah I'm not sure why I did this. I remember at one point encountering a compiler error that I was surprised by; maybe it was related to this. I'll take a look.

core/src/main/scala/cats/collections/Set.scala
ceedubs: I wasn't sure about this either. I'm happy to undo it. I think that this means removing the covariance from the `V` in `AvlMap` as well.

ceedubs:
Thanks for the the review, @johnynek! You bring up good points. I'll revisit these items.

core/src/main/scala/cats/collections/Set.scala
ceedubs: I guess that this same argument would apply to `Diet`, wouldn't it? `Diet` is the structure that I was originally wanting to be covariant, but now I'm questioning this decision. I guess that really all I want is to be able to say is that `Empty` is universal over `A`. It's a bummer that this isn't straightforward in Scala.

ceedubs:
Prompted by @johnynek's comments, I'm starting to think that maybe these changes only actually make sense for `Range` and `Dequeue` where the construction and usage of the structure doesn't depend on consistent `Order` instances throughout the program. Correctness aside, the covariance does introduce quite a bit of noise in the structures that take implicit type class instances for most/all methods. WDYT, @larsrh and @johnynek?

core/src/main/scala/cats/collections/Set.scala
larsrh: The more ambitious change, if you'd be up for it, would be to define `AvlSet` in terms of `AvlMap`, with `V = Unit`.

ceedubs: I'm not sure if I'm that ambitious 😬

larsrh:
> I'm starting to think that maybe these changes only actually make sense for `Range` and `Dequeue` where the construction and usage of the structure doesn't depend on consistent `Order` instances throughout the program.

Hmm, good question. I have no strong opinion either way. I'll post this in typelevel/general to get some opinions.

jdegoes:
@ceedubs Excellent work and much needed!!! 🕺 🕺 🕺 

As noted, this will in general break source compatibility.

ceedubs:
I'm going to close this out and submit some individual PRs to decouple the decisions on the different structures.

PR #213, opened 2019-06-29
Add one and fromRange methods for creating Diet

ceedubs:
Since the `Diet` constructors are package private, these methods can't be written in their current form outside of `cats.collections`. You can mimic them by calling `Diet.empty` and adding elements to it, but it is a little more verbose (and it applies some unnecessary `Order` and `Discrete` constraints).

### DIFF
--- a/core/src/main/scala/cats/collections/Diet.scala
+++ b/core/src/main/scala/cats/collections/Diet.scala
@@ -277,6 +277,16 @@ sealed abstract class Diet[A] {
 object Diet {
   def empty[A]: Diet[A] = EmptyDiet()
 
+  /**
+   * Create a Diet that contains only a single `A` element.
+   */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/collections/Diet.scala
```diff
@@ -277,6 +277,16 @@ sealed abstract class Diet[A] {
 object Diet {
   def empty[A]: Diet[A] = EmptyDiet()
 
+  /**
+   * Create a Diet that contains only a single `A` element.
+   */
+  def one[A](a: A): Diet[A] = DietNode(Range(a, a), empty, empty)
```
ceedubs: I chose `one` to be consistent with `NonEmptyList.one`.

Issue #193, opened 2019-05-12
Heap.pop()?


ryan-williams:
Any reason that `Heap.remove` doesn't return the min element in addition to the new Heap, i.e. `(Option[A], Heap[A])`?

I'm new to this repo, but would have guessed that was the much more common operation.

Perhaps a new `pop` method would be the way to add it?

This is simple and I need it, so I'll PR it if it sounds good, thanks!

johnynek:
`heap.minimumOption.map { m => (m, heap.remove) }` would do it.

It would be a nice PR.

ryan-williams:
Thanks. Thinking out loud for a second because I may be confused about what I need.

My plan is to use Heap in a cats-effect `Ref` and have multiple workers concurrently pulling work-items off of it.

I thought calling `minimumOption` and `remove` in sequence would introduce a race condition, but I guess the idea is that doing them in a single `Ref.modify` guarantees that the `Heap` was not modified during the sequence.

So that answers my initial question ("Any reason…"), but nevertheless we're saying that a `pop` helper that combines both would be nice.

Let me know if anything jumps out as wrong there? Thanks again.

johnynek:
That sounds right to me. Modify is atomic.

anicolaspp:
@ryan-williams the reason is that I didn't think about it in my original implementation. Initially, there was Stew and me, we didn't have a lot of visibility and we were trying to push the library out there, so we didn't have a lot of requirements, just trying to do something that could help others, and here we are...

cb372:
This issue can be closed, right? It’s been implemented.

johnynek:
done.

https://github.com/typelevel/cats-collections/blob/361e6b0c55a14969c8119934b60643bb30fc6f33/core/src/main/scala/cats/collections/Heap.scala#L150

https://github.com/typelevel/cats-collections/blob/361e6b0c55a14969c8119934b60643bb30fc6f33/core/src/main/scala/cats/collections/PairingHeap.scala#L176

https://github.com/typelevel/cats-collections/blob/361e6b0c55a14969c8119934b60643bb30fc6f33/core/src/main/scala/cats/collections/PartiallyOrderedSet.scala#L138

PR #216, opened 2019-07-08
Implement Heap.pop

erwan:
To fix this: https://github.com/typelevel/cats-collections/issues/193

I implemented `Heap.pop`. Since `Heap.remove` can be derived from `Heap.pop`, I changed the implementation of `remove` to reuse `pop`.

### DIFF
--- a/core/src/main/scala/cats/collections/Heap.scala
+++ b/core/src/main/scala/cats/collections/Heap.scala
@@ -130,6 +130,15 @@ sealed abstract class Heap[A] {
     Heap.heapify(a)
 
   /**
+   * Remove the min element from the heap (the root) and return it along with the updated heap.
+   * Order O(log n)
+   */
+  def pop(implicit order: Order[A]): Option[(A, Heap[A])] = this match {
+    case Branch(m, l, r) => Some((m, bubbleRootDown(mergeChildren(l, r))))
+    case Leaf()          => None
+  }
+
+  /**
    * Remove the min element from the heap (the root).
    * Order O(log n)
    */
--- a/core/src/main/scala/cats/collections/PairingHeap.scala
+++ b/core/src/main/scala/cats/collections/PairingHeap.scala
@@ -161,6 +161,15 @@ sealed abstract class PairingHeap[A] {
   }
 
   /**
+   * Remove the min element from the heap (the root) and return it along with the updated heap.
+   * Order O(log n)
+   */
+  def pop(implicit order: Order[A]): Option[(A, PairingHeap[A])] = this match {
+    case Tree(m, subtrees) => Some((m, combineAll(subtrees)))
+    case Leaf()            => None
+  }
+
+  /**
    * if not empty, remove the min, else return empty
    * this is thought to be O(log N) (but not proven to be so)
    */
--- a/tests/src/test/scala/cats/collections/HeapSpec.scala
+++ b/tests/src/test/scala/cats/collections/HeapSpec.scala
@@ -120,6 +120,24 @@ class HeapSpec extends CatsSuite {
     assert(Heap.empty[Int].remove == Heap.empty[Int])
   }
 
+  test("pop and remove return the same heap") {
+    forAll { (heap: Heap[Int]) =>
+      val heap1 = heap.remove
+      heap.pop.map(_._2) match {
+        case Some(heap2) => assert(Order[Heap[Int]].eqv(heap1, heap2))
+        case None => assert(heap1.isEmpty)
+      }
+    }
+  }
+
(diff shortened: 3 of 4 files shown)


core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -130,14 +130,20 @@ sealed abstract class Heap[A] {
...
+  def pop(implicit order: Order[A]): (Heap[A], Option[A]) = this match {
+    case Branch(m, l, r) => (bubbleRootDown(mergeChildren(l, r)), Some(m))
+    case Leaf()          => (this, None)
   }
 
+  /**
+   * Remove the min element from the heap (the root).
+   * Order O(log n)
+   */
+  def remove(implicit order: Order[A]): Heap[A] = this.pop._1
```
johnynek: Can we keep the old implementation to save some allocations on this variant?  I’m less interested in code reuse than performance with libraries we hope to be adopted widely.

core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -130,14 +130,20 @@ sealed abstract class Heap[A] {
...
 
   /**
-   * Remove the min element from the heap (the root).
+   * Remove the min element from the heap (the root) and return it along with the updated heap.
    * Order O(log n)
    */
-  def remove(implicit order: Order[A]): Heap[A] = this match {
-    case Branch(_, l, r) => bubbleRootDown(mergeChildren(l, r))
-    case Leaf()          => Leaf()
+  def pop(implicit order: Order[A]): (Heap[A], Option[A]) = this match {
```
johnynek: I think `Option[(A, Heap[A])]` would be a better signature here, and I think only fewer allocations (since the empty case doesn’t allocate at all) and the non empty case allocated one tuple and one Some just as your implementation does.

johnynek:
Also I think we should push this API to Heaplike and PairingHeap as well.

johnynek:
Actually we changed HeapLike to PartiallyOrderedSet

core/src/main/scala/cats/collections/Heap.scala
erwan: ok 👍

erwan:
`PartiallyOrderedSet` seems to have different naming for its methods, and it already have an `unadd` method that seems to be doing that

core/src/main/scala/cats/collections/Heap.scala
erwan: ok

erwan:
@johnynek I did the changes except for `PartiallyOrderedSet`, because I think the method is already there as `unadd`

larsrh:
There's currently a build problem, caused by the last release. @djspiewak are you aware of this?

djspiewak:
@larsrh I'll take a look later today.

johnynek:
one last question on naming:

I would say `push/pop` are nice pairs. But `add/pop` not as much. That's why I used `unadd` which is a bit strange, but kind of clearly the dual of add.

Also, the typeclass is using `unadd`. It seems like we should either change that one, or use unadd here.

What do you think?

erwan:
I'm not really used to `unadd` actually, I think it's the first time I see it.

I'm used to push/pop when the elements are always added or removed at the end, and add/remove when it's in any position.

If it was me I would just use push/pop everywhere, unless there is a reason to use add/unadd in this case?

larsrh:
I looked in other languages (e.g. Haskell calls it `view`, `viewHead` and `viewTail`, OCaml `push` and `pop`), so no clear consensus there. In the end I think `push` and `pop` sound reasonable.

Plan: deprecate the old name, introduce the new ones everywhere. I'm happy to have that in a separate PR, though; however you prefer @erwan.

erwan:
@larsrh sounds good, I'll add a commit for that in this PR if it's OK for you

erwan:
hmmm, actually we can't really deprecate anything (as in mark it with the `@deprecated` annotation) because when we use it in tests it generates a warning, that `-Xfatal-warning` turns into a compilation error...

core/src/main/scala/cats/collections/Heap.scala
anicolaspp: I originally implemented `heap` with performance and immutability in mind, be careful with changes that change those properties.

larsrh:
@erwan We can get rid of that flag, no problem.

larsrh:
@erwan #217 is merged, could you please rebase this one on top?

erwan:
@larsrh done!

larsrh:
Thanks!

PR #242, opened 2019-10-05
Update cats to 2.0.0 version

wojciechUrbanski:
After upgrade to cats 2.0.0 BitSet tests were failing with NoClassDefFound. In order to fix the problem, I have rewritten the BitSet so it uses the CatsSuite instead of Property from Scalacheck. To keep the consistency with previous settings, I have set up the config to run at least 1000 successful tests.


build.sbt
```diff
@@ -91,7 +91,8 @@ lazy val tests = crossProject(JSPlatform, JVMPlatform)
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-testkit" % V.cats % "test",
       "org.typelevel" %%% "cats-laws"    % V.cats % "test",
-      "org.typelevel" %%% "algebra-laws" % V.algebra % "test"
+      "org.typelevel" %%% "algebra-laws" % V.algebra % "test",
+      "org.typelevel" %% "cats-testkit-scalatest" % V.catsTestkit % "test"
```
larsrh: I'm a little surprised why this works, because I thought a `%%%` is required here. :thinking:

larsrh:
We should probably also wait until typelevel/algebra#234, which I'll try to get done this weekend. (Not that that changes anything in your PR, but it's just good dependency hygiene.)

build.sbt
wojciechUrbanski: Wow, i was looking everywhere for some kind of a mistake due to NoClassDefFound but missed that one... Thanks for pointing it out :) I will add the `%%%` and revert changes from BitSetTest - it is working just fine.

PR #247, opened 2019-10-25
Diet eq

anicolaspp:


core/src/main/scala/cats/collections/Diet.scala
```diff
@@ -233,10 +234,18 @@ sealed abstract class Diet[A] {
     case DietNode(_, _, r) => r.max
   }
 
-  @deprecated("the semantics of this method are unclear and unspecified, avoid!", "0.8.0")
-  def map[B: Discrete: Order](f: A => B): Diet[B] = this match {
+  //  @deprecated("the semantics of this method are unclear and unspecified, avoid!", "0.8.0")
```
larsrh: I strongly disagree with un-deprecating this. Mapping may destroy the invariants.

tests/src/test/scala/cats/collections/DietSpec.scala
```diff
@@ -101,6 +101,13 @@ class DietSpec extends CatsSuite {
     (Diet.empty[Int] - Range(10, 100)) should be (Diet.EmptyDiet())
   }
 
+  test("diet eq") {
```
larsrh: This demands at least one or two Scalacheck tests.

anicolaspp: I will add some of those

larsrh:
Thanks a lot!

anicolaspp:
I am happy to be back

PR #249, opened 2019-10-28
Range.toIterator

anicolaspp:


core/src/main/scala/cats/collections/Range.scala
```diff
@@ -12,96 +12,83 @@ final case class Range[A](val start: A, val end: A) {
...
+   * start to end taking into consideration the provided ordering.
+   * If (start > end) it uses start's predecessor offered by the Discrete
+   * instance, otherwise it uses the start's successor.
+   */
+  def toIterator(implicit discrete: Discrete[A], order: Order[A]): Iterator[A] = {
+    val next: A => A = a => if (order.lteqv(a, end)) discrete.succ(a) else discrete.pred(a)
+
+    Stream
+      .iterate(start)(next)
+      .takeWhile(concurrent => !order.eqv(concurrent, end))
```
johnynek: Why was this named concurrent rather than current?

PR #252, opened 2019-10-29
Trying to fix the history

anicolaspp:


larsrh:
I think you have to reset your local `master` (or fork again).

anicolaspp:
@larsrh for some reason I cannot remove the old commits. It seems I need to PR from different branches which was my mistake. If you could review and merge this I will update my master and then continue to PR future work from a different branch

anicolaspp:
@larsrh let me fork again. That should fix it

PR #253, opened 2019-10-29
heap.toIterator

anicolaspp:


core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -129,6 +129,19 @@ sealed abstract class Heap[A] {
   def heapify(a: List[A])(implicit order: Order[A]): Heap[A] =
     Heap.heapify(a)
 
+  def toIterator(implicit order: Order[A]): Iterator[A] = {
+    @scala.annotation.tailrec
+    def build(heap: Heap[A], acc: Stream[A]): Stream[A] = heap match {
```
johnynek: Why is using a Stream better than just building the List below? Since it is tailrec you are building the whole Stream no? If we do that, isn’t a List cheaper?

anicolaspp: in list you have to reverse it at the end.

anicolaspp: In this particular case the following does not apply, but in the other implementations, the stream based removed the need for mutations, and adds stateless build of the iterator.

johnynek: If we are making a performance argument, can we show a win with benchmarks?

anicolaspp: Oh man, you are killing me... I haven't written one of those in a while. Let me see

Issue #202, opened 2019-06-18
Clean up `Predicate`


larsrh:
According to @johnynek's review comments:

https://github.com/typelevel/cats-collections/pull/201#pullrequestreview-250248153

PR #269, opened 2020-01-16
Clean Up Predicate A Bit

stephen-lazaro:
Alias existing functions from symbolic ones
Add contravariant instance
Guarantee all instances serializable

For #202.

### DIFF
--- a/core/src/main/scala/cats/collections/Predicate.scala
+++ b/core/src/main/scala/cats/collections/Predicate.scala
@@ -16,7 +16,7 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
   /**
    * returns a predicate which is the union of this predicate and another
    */
-  def |[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) || other(a))
+  def |[B <: A](other: Predicate[B]): Predicate[B] = self union other
 
   /**
    * returns a predicate which is the intersection of this predicate and another
@@ -26,7 +26,7 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
   /**
    * returns a predicate which is the intersection of this predicate and another
    */
-  def &[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) && other(a))
+  def &[B <: A](other: Predicate[B]): Predicate[B] = self intersection other
 
   /**
    * Returns true if the value satisfies the predicate.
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -63,12 +63,20 @@ object Predicate extends PredicateInstances {
...
+    override def unit: Predicate[Unit] = Predicate.empty
+  }
+
   implicit def predicateMonoid[A]: Monoid[Predicate[A]] = new Monoid[Predicate[A]] {
     override def empty: Predicate[A] = Predicate.empty
     override def combine(l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
   }
 
-  implicit val predicateInstance: MonoidK[Predicate] = new MonoidK[Predicate] {
+  implicit val predicateMonoidK: MonoidK[Predicate] = new MonoidK[Predicate] {
```
stephen-lazaro: Dunno if this rename is cool from a bincompat perspective.

larsrh: There are no guarantees about that yet, so it's fine.

johnynek: I think we should add mima checking sooner than later.

I personally really avoid libraries that don't try to minimize breaks (and in fact, despite contributing to this library, I am afraid to depend on it).

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -63,12 +63,20 @@ object Predicate extends PredicateInstances {
 }
 
 trait PredicateInstances {
+  implicit def predicateContravariantMonoidal: ContravariantMonoidal[Predicate] = new ContravariantMonoidal[Predicate] {
+    override def contramap[A, B](fb: Predicate[A])(f: B => A): Predicate[B] =
+      Predicate(f andThen fb.apply)
+    override def product[A, B](fa: Predicate[A], fb: Predicate[B]): Predicate[(A, B)] =
+      Predicate(v => fa(v._1) || fb(v._2))
```
larsrh: I'm curious about the choice of `||` here, what's the reasoning?

core/src/main/scala/cats/collections/Predicate.scala
larsrh: Happy to call it 1.0 at some point in the near future.

core/src/main/scala/cats/collections/Predicate.scala
stephen-lazaro: Yeah, it's definitely a bit arbitrary, but I made the choice based on having the monoidal product act similarly to the monoids already defined for this type. `||` gives us union, which is probably the more in line with user expectations given the monoids are also union.

larsrh: Could you please add law tests for `ContravariantMonoidal`? Then I think we're good to go.

stephen-lazaro: Sure, I'll get to that today / tomorrow probably.

stephen-lazaro: I did it with kind of stupid `Eq` instances, but I'm not sure there are better ones. If there's a standard way to do `Eq` for functions, let me know and I can change it up.

larsrh:
Thanks @stephen-lazaro!

Issue #142, opened 2018-11-04
A purely functional Set implementation using `Eq`


nasadorian:
I noticed that the distinct implementations in cats data types are using on Scala's `TreeSet` and the `Order` typeclass. But what about a purely functional set based on `Eq` using only boolean logic?

Upsides:

* Intuitively more performant than TreeSet, with O(1) lookup and insert
* Small memory footprint per object, so far just a case class
* Uses Eq instead of Order or Hash, which is closest to the mathematical definition of Set semantics 
* Could give rise to a Predicate data type for cats as well, similar to the one seen in Haskell docs (http://hackage.haskell.org/package/contravariant-1.5/docs/Data-Functor-Contravariant.html)

Downsides:

* Ironically, could not derive an Eq instance for the set itself since it's based on a function?
* GC pressure due to copying of case classes
* No way to inspect elements, since it is only optimized for checking containment
* No way to check size

I wrote up a basic implementation (https://gist.github.com/nasadorian/8f10cc1f89225a932238026cf46e57a7) using just a case class and `Eq`. 

Would like to improve it and use more lawful constructs - I am thinking a better impl could depend on some Contravariant Functor like `Predicate` since it is "containing" a function, but I'm not totally sure how to make that work. 

The Set itself could have Monoid and Semigroup instances easily. I would also like to figure out if it's possible to make it a Functor.

Cheers!

larsrh:
That should basically be [ISet](https://github.com/typelevel/cats-collections/blob/master/core/src/main/scala/cats/collections/ISet.scala). Making it a functor [wouldn't work,](https://typelevel.org/blog/2014/06/22/mapping-sets.html) however.

nasadorian:
Thanks @larsrh - looks like I'm a few years late to the party 😅. I do have some improvements I think I could make in a PR.

ISet seems to be based on a generic predicate. Would the special case based on Eq (which I'd assume is most common anyway) belong in the same library?

nasadorian:
Also interestingly, it seems neither ISet nor my implementation are not stack safe after a certain size. Found that ISet stack overflows when looking up in a set of around 20k elements.

larsrh:
It depends on how you construct such a set. One possibility to make it stack-safe would be to use a function `A => cats.Eval[Boolean]` instead of `A => Boolean`.

LukaJCB:
I think `Eval[A => Boolean]` would be even better (similar to what we did for `State`) as function composition isn't stack safe in general either

[...]

PR #283, opened 2020-03-05
Make Predicate stack-safe using Eval

bbjubjub2494:
Fixes last comment in #142, Also adds `contramap()` ~~to compensate for the loss of inheritance from `Function1`.~~

# Benchmarks
## pre-existing 0.9.0 stack-unsafe implementation
(at 992770b8685ee0287df19b068a4d1f14818d27b7)
```
Benchmark                                               (n)   Mode  Cnt        Score       Error  Units
ChainedPredicateBench.catsCollectionsPredicateUnravel    10  thrpt   25  8168148.133 ± 83522.583  ops/s
ChainedPredicateBench.catsCollectionsPredicateUnravel   100  thrpt   25   635545.444 ±  3674.601  ops/s
ChainedPredicateBench.catsCollectionsPredicateUnravel  1000  thrpt   25    51776.603 ±   335.972  ops/s
```
(SOE at n = 10,000)

## this PR (mark 3)
```
Benchmark                                                (n)   Mode  Cnt        Score       Error  Units
ChainedPredicateBench.catsCollectionsPredicateUnravel     10  thrpt   25  3092451.360 ± 20246.653  ops/s
ChainedPredicateBench.catsCollectionsPredicateUnravel    100  thrpt   25   311739.000 ±  3842.162  ops/s
ChainedPredicateBench.catsCollectionsPredicateUnravel   1000  thrpt   25    27788.238 ±   226.262  ops/s
ChainedPredicateBench.catsCollectionsPredicateUnravel  10000  thrpt   25     2626.175 ±    20.365  ops/s
```
That's roughly -50% at n = 100 and -45% at n = 1,000. It was a convoluted intensive example with all the stress on the internals, so I would consider it a worst case scenario. Most real world intensional sets probably put most of their complexity in the basic predicates rather than in thousands of combinations. Hence I would say -50% is the lower bound of the performance degradation in general. Is that acceptable?

### DIFF
--- a/core/src/main/scala/cats/collections/Predicate.scala
+++ b/core/src/main/scala/cats/collections/Predicate.scala
@@ -132,7 +132,7 @@ object Predicate extends PredicateInstances {
    * build a set from a membership function.
   */
   def apply[A](p: A => Boolean): Predicate[A] = Lift {
-    Kleisli(a => Eval.now(p(a)))
+    Kleisli(a => if (p(a)) Eval.True else Eval.False)
   }
 
   /**
--- a/core/src/main/scala/cats/collections/Predicate.scala
+++ b/core/src/main/scala/cats/collections/Predicate.scala
@@ -147,7 +147,7 @@ object Predicate extends PredicateInstances {
 }
 
 trait PredicateInstances {
-  implicit def predicateContravariantMonoidal: ContravariantMonoidal[Predicate] = new ContravariantMonoidal[Predicate] {
+  implicit def catsCollectionsPredicateContravariantMonoidal: ContravariantMonoidal[Predicate] = new ContravariantMonoidal[Predicate] {
     override def contramap[A, B](fb: Predicate[A])(f: B => A): Predicate[B] =
       fb.contramap(f)
     override def product[A, B](fa: Predicate[A], fb: Predicate[B]): Predicate[(A, B)] =
@@ -155,12 +155,12 @@ trait PredicateInstances {
     override def unit: Predicate[Unit] = Predicate.empty
   }
 
-  implicit def predicateMonoid[A]: Monoid[Predicate[A]] = new Monoid[Predicate[A]] {
+  implicit def catsCollectionsPredicateMonoid[A]: Monoid[Predicate[A]] = new Monoid[Predicate[A]] {
     override def empty: Predicate[A] = Predicate.empty
     override def combine(l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
   }
 
-  implicit def predicateBool[A]: Bool[Predicate[A]] = new Bool[Predicate[A]] {
+  implicit def catsCollectionsPredicateBool[A]: Bool[Predicate[A]] = new Bool[Predicate[A]] {
     override def one: Predicate[A] = Predicate.everything
     override def zero: Predicate[A] = Predicate.empty
     override def complement(x: Predicate[A]): Predicate[A] = x.negate
@@ -169,7 +169,7 @@ trait PredicateInstances {
 
   }
 
-  implicit val predicateMonoidK: MonoidK[Predicate] = new MonoidK[Predicate] {
+  implicit val catsCollectionsPredicateMonoidK: MonoidK[Predicate] = new MonoidK[Predicate] {
     override def empty[A]: Predicate[A] = Predicate.empty
     override def combineK[A](l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
   }
--- a/core/src/main/scala/cats/collections/Predicate.scala
+++ b/core/src/main/scala/cats/collections/Predicate.scala
@@ -1,5 +1,6 @@
 package cats.collections
 
+import algebra.lattice.Bool
 import cats._
 import cats.data.Kleisli
 
@@ -157,6 +158,15 @@ trait PredicateInstances {
   implicit def predicateMonoid[A]: Monoid[Predicate[A]] = new Monoid[Predicate[A]] {
     override def empty: Predicate[A] = Predicate.empty
     override def combine(l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
+  }
+
+  implicit def predicateBool[A]: Bool[Predicate[A]] = new Bool[Predicate[A]] {
+    override def one: Predicate[A] = Predicate.everything
+    override def zero: Predicate[A] = Predicate.empty
+    override def complement(x: Predicate[A]): Predicate[A] = x.negate
+    override def and(l: Predicate[A], r: Predicate[A]): Predicate[A] = l intersection r
+    override def or(l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
+
   }
 
   implicit val predicateMonoidK: MonoidK[Predicate] = new MonoidK[Predicate] {
--- a/core/src/main/scala/cats/collections/Predicate.scala
+++ b/core/src/main/scala/cats/collections/Predicate.scala
@@ -1,17 +1,39 @@
 package cats.collections
 
 import cats._
+import cats.data.Kleisli
 
 /**
  * An intensional set, which is a set which instead of enumerating its
  * elements as a extensional set does, it is defined by a predicate
  * which is a test for membership.
+ *
+ * All combinators in this class are implemented in a stack-safe way.
  */
-abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
+sealed abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
+  def run: Kleisli[Eval, A, Boolean]
+
+  /**
+   * Returns true if the value satisfies the predicate.
+   */
+  def apply(a: A): Boolean = run(a).value
+
   /**
    * returns a predicate which is the union of this predicate and another
    */
-  def union[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) || other(a))
+  def union[B <: A](other: Predicate[B]): Predicate[B] =
+    self match {
+      case Predicate.Empty => other
+      case Predicate.Everything => self
+      case _ =>
+      other match {
+        case Predicate.Empty => self
+        case Predicate.Everything => other
+        case _ => Predicate.Lift {
+          self.run.flatMap(if (_) Predicate.True else other.run)
+        }
+      }
+    }
 
   /**
    * returns a predicate which is the union of this predicate and another
@@ -21,7 +43,19 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
   /**
    * returns a predicate which is the intersection of this predicate and another
    */
-  def intersection[B <: A](other: Predicate[B]): Predicate[B] = Predicate(a => apply(a) && other(a))
+  def intersection[B <: A](other: Predicate[B]): Predicate[B] =
+    self match {
+      case Predicate.Empty => self
+      case Predicate.Everything => other
+      case _ =>
+        other match {
+        case Predicate.Empty => other
+        case Predicate.Everything => self
+        case _ => Predicate.Lift {
+          self.run.flatMap(if (_) other.run else Predicate.False)
+        }
+      }
+    }
 
   /**
    * returns a predicate which is the intersection of this predicate and another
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -2,16 +2,35 @@ package cats.collections
 
 import cats._
 
+private object PredicateHelpers {
+  // use the static Eval instances to reduce memory footprint
+  @inline def boolToNow(b: Boolean) = if (b) Eval.True else Eval.False
+
+  trait FromApplyF[A] extends Predicate[A] {
+    override def applyF(a: A): Eval[Boolean] =
+      throw new NotImplementedError("applyF implementation missing")
```
johnynek: is this needed? Why not just leave it abstract? I guess because otherwise you have an infinite loop and you want a cleaner error? If so, can you add a comment to that effect?

lourkeur: It seems it cant be made to be abstract again because it is concrete in Predicate. I tried `abstract override` too but to no avail.

lourkeur:
> I'm concerned with the drop in performance here. Can you make a benchmark?
> 
> There are other ways to make Predicate stack safe without giving up all the performance (see what was done in cats): https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/AndThen.scala

Will get back to you on that. I think a mix between using AndThen where possible and Eval otherwise might work.

lourkeur:
I'm thinking the following regarding API:
- don't inherit from Function1: we can't inherit from AndThen and if we just have Function1, we could have instances where the AndThen optimized composition wouldn't kick in silently. Is `def apply: AndThen[A, Boolean]` a good idea?
- `compose` and `andThen` are staying, we're proxying AndThen's interface essentially.
- Should AndThen be visible in the interface? It doesn't need to be so maybe it's more flexible that way.
- We should expose 2 factory functions: one from `A => Boolean` and one from `A => Eval[Boolean]` and also use them internally. According to [cats](https://github.com/typelevel/cats/blob/fc230798d830106198b2b34cd9d93b1478c55612/core/src/main/scala/cats/data/ContT.scala#L52), 2 subclasses is the magic number so I'll aim for that.
- `applyEval` should be exposed as well I think.

lourkeur:
I believe this is not stack-safe under something like repeated `_.negate union Empty` because we're having to call `.value`.
- [] add a regression test against that
- [] move to `AndThen[A, Eval[Boolean]]`, then test
- [] benchmark that implementation against `Eval.Defer`

lourkeur:
The PR stabilized and finalized. @johnynek would you mind reviewing again or pinging somebody else?

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -46,28 +60,62 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
...
+    case e => e.flatMap { if (_) Eval.False else Eval.True }
+  })
 
   /**
    * Return the opposite predicate
    */
   def unary_!(): Predicate[A] = negate
+
+  /**
+   * compose with a function the predicate.
```
lourkeur, line 74:
```suggestion
   * compose a function with the predicate.
```
typo on my part, will fix promptly

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -1,17 +1,29 @@
...
 
 /**
  * An intensional set, which is a set which instead of enumerating its
  * elements as a extensional set does, it is defined by a predicate
  * which is a test for membership.
+ *
+ * All operations in this class are implemented in a stack-safe way.
  */
-abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
+final class Predicate[-A] private(private val p: AndThen[A, Eval[Boolean]]) { self =>
```
johnynek, line 13: I'm still pretty skeptical of the need for both AndThen and Eval. Eval comes with a big performance penalty, like 100x isn't uncommon.

Is blowing the stack very common here?

Next, Why not just make `type Predicate[-A] = Kleisli[Eval, A, Boolean]` and then have some methods on that if you want to have a stack safe predicate? Why a new class?

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -46,28 +60,62 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
   /**
    * Return the opposite predicate
    */
-  def negate: Predicate[A] = Predicate(a => !apply(a))
+  def negate: Predicate[A] = Predicate.wrap[A](p.andThen {
+    case e: Now[Boolean] => if (e.value) Eval.False else Eval.True
+    case e => e.flatMap { if (_) Eval.False else Eval.True }
+  })
```
johnynek, line 66: if we had case classes you could defer the code here:
```scala
case class Negate[A](of: Predicate[A]) extends Predicate[A] {
 ...
}
```
then `def negate` could apply the law `negate(negate(p)) == p`. Similar with keeping `And`, `Or` and `Const` nodes around.

core/src/main/scala/cats/collections/Predicate.scala
lourkeur: Regarding performance, I'm not seeing any 100x: I posted a benchmark in the opening post and it seems to be at 2x worst case scenario. I might be missing something though.

Regarding the need for `Eval`, since we are dealing with tree-shaped expressions, there is a need for some type of trampoline like `Eval` if stack safety is to be achieved. (A `Defer[_]` to be precise) Maybe `scala.util.control.TailCalls.TailRec` would be preferable since we dont need all the laziness modes?

Regarding `AndThen`, it makes contramap stack-safe, and it's a thin layer over functions that doesn't even kick in unless a lot of composition is done, so I think it's actually more performant in this role than Kleisli because Kleisli leans on the trampoline. I'll come with benchmarks though.

Regarding whether it's common to blow the stack here, I come at it from the perspective that it shouldn't be possible at all in a Cats library. It should be as efficient as possible while remaining SOE-less. That's my interpretation of the "Efficency" motivation on the [Cats page](https://typelevel.org/cats/motivations.html#efficiency).

Regarding the class question that's a good point. I tried turning the class into an opaque newtype a couple of months ago, but it wouldn't compile under 2.11. Now that 2.11 support has been dropped we could go that route, or do a transparent type alias.

core/src/main/scala/cats/collections/Predicate.scala
lourkeur: I'm not really in favor of adding an AST layer.

I feel like most of the time these kinds of optimisations will be spotted by user code that also knows these laws, and that most of the nodes will be opaque predicates (`A => Boolean`) that we cant really do a lot with besides double negation. So the laws would rarely be applied in practice. This also clashes with the type alias route, so simple predicates get an additional indirection, and adds a bunch of boilerplate and complexity.

Side note: would we apply De Morgan's laws in either direction?

core/src/main/scala/cats/collections/Predicate.scala
lourkeur: Regarding trampolines, I realized the most suitable trampoline is probably a dedicated one: we only use Eval.{True, False}, defer, and flatMap, so rolling my own should be fairly easy and would have built-in unboxed booleans. We shall see on the benchmarks.

johnynek: Sorry, I overlooked the benchmarks, those look good to me. My concerns are removed by those.

I don't think a custom trampoline can speed this up (maybe I'm wrong) since there are only two Booleans boxing them is almost free, and `Eval.True` and `Eval.False` have already boxed them. Eval is faster than scala TailCall and cats.free.Free (by a significant margin) so beating it seems tough to me.

lourkeur: I'm now seeing a combination of the trampoline and the AST, similar to a trampoline with a monadReader capability. It's probably a tiny bit faster, it's more elegant and allows to apply laws so I'll try that next. Thanks you for stimulating my thought process.

lourkeur:
Ok there we go. Kleisli is much nicer to work with. We have AST simplification, a more complete scalacheck generator, and the Boolean algebra. I haven't run final benchmarks yet but I think performance profile is the same as 2 days ago.

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -46,28 +81,77 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
...
+  private[collections] final case class Negate[A](negate: Predicate[A]) extends Predicate[A] {
+    override def run: Kleisli[Eval, A, Boolean] = negate.run.map(!_)
+    override def contramap[B](f: B => A): Predicate[B] = Negate(negate contramap f)
+  }
+
+  /**
+   * build a set from a membership function.
+  */
+  def apply[A](p: A => Boolean): Predicate[A] = Lift {
+    Kleisli(a => Eval.now(p(a)))
```
johnynek, line 135: I think `a => if (p(a)) Eval.True else Eval.False` may be marginally more efficient.

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -76,6 +160,15 @@ trait PredicateInstances {
     override def combine(l: Predicate[A], r: Predicate[A]): Predicate[A] = l union r
   }
 
+  implicit def predicateBool[A]: Bool[Predicate[A]] = new Bool[Predicate[A]] {
```
johnynek, line 163: we usually give longer names in cats to avoid the name aliasing problem with implicits.

e.g. `implicit def catsCollectionsPredicateBool: Bool[Predicate[A]] = ...`

tests/src/test/scala/cats/collections/PredicateSpec.scala
```diff
@@ -22,6 +24,16 @@ class PredicateSpec extends CatsSuite {
     checkAll("ContravariantMonoidal[Predicate]", ContravariantMonoidalTests[Predicate].contravariantMonoidal[Int, Int, Int])
   }
 
+  {
+    implicit val eqForPredicateInt: Eq[Predicate[Int]] = new Eq[Predicate[Int]] {
+      val sample = -1 to 1 // need at least 2 elements to distinguish in-between values
```
johnynek, line 29: why are just two enough? Why not `(-100 to 100)` or something?

Or, better yet, why not test with `Predicate[Byte]` and enumerate all `256` possibilities to check equality.

lourkeur: If the domain is just one element, then every predicate is effectively equivalent to either Empty or Everything depending on whether it accepts the one element.

I don't think we need a 201 or 256 cardinality domain because I tried manually introducing some defects in Predicate and it caught them immediately with the 3 elements.

That being said, it doesn't negatively affect the performance of the testsuite on my machine at all so I'd be ok with either.

lourkeur:
# Benchmarks
## tip of PR
```
[info] Benchmark                                                (n)   Mode  Cnt        Score       Error  Units
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel     10  thrpt   25  1441476.494 ± 16104.839  ops/s
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel    100  thrpt   25   150259.432 ±   447.757  ops/s
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel   1000  thrpt   25    15139.120 ±    36.946  ops/s
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel  10000  thrpt   25     1509.519 ±    15.303  ops/s
```

## same benchmark against on master
(currently running)

lourkeur:
# benchmark without Kleisli
```
[info] Benchmark                                                (n)   Mode  Cnt        Score       Error  Units
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel     10  thrpt   25  4396333.243 ± 64479.117  ops/s
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel    100  thrpt   25   520604.540 ±  2450.074  ops/s
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel   1000  thrpt   25    48156.396 ±   125.049  ops/s
[info] ChainedPredicateBench.catsCollectionsPredicateUnravel  10000  thrpt   25     4519.274 ±     7.830  ops/s
```
still 5x

core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -46,37 +81,95 @@ abstract class Predicate[-A] extends scala.Function1[A, Boolean] { self =>
...
+    override def contramap[B](f: B => A): Predicate[B] = Negate(negate contramap f)
+  }
+
+  /**
+   * build a set from a membership function.
+  */
+  def apply[A](p: A => Boolean): Predicate[A] = Lift {
+    Kleisli(a => if (p(a)) Eval.True else Eval.False)
+  }
+
```
johnynek, line 137: what about a `def fromKleisli[A](k: Kleisli[Eval, A, Boolean]): Predicate[A] = Lift(k)`

johnynek (commented):
one small comment

Can we remove the draft? I think this is close to ready for merge (even could be merged now, and my suggestion could be added before publishing).

PR #304, opened 2020-07-08
HashMap based on Eq and Object#hashCode

s5bug:
For when there's an `Eq` with no `Order`. I'm not the best at designing HashMaps nor cats-centric collections; I'm opening as a draft PR in hopes of live feedback.


sorenbug:
How do I add tests? `SetSpec` refers specifically to `AvlSet` (same for `MapSpec`/`AvlMap`).

larsrh:
Hi @sorenbug and thanks for your contribution! The current set/map files should really be prefixed with `Avl` (this includes the tests).

As for your contribution, I recommend putting them into separate files, such as `HashMap.scala`. I can send a follow-up PR that renames the existing files.

To write tests, I think you could largely copy the existing tests and change occurrences of `AvlSet` to your `HashSet`.

sorenbug:
Sounds good! I will do so.

johnynek:
Also, @sorenbug I think we'd want to require `Hash[K]` not `Eq[K]`:

https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/Hash.scala

We don't want to use universal hashing `.hashCode` in a typeclass based library like this one.

Lastly, getting a high performance HashMap will be a bit tricky, but I think directly implementing this via the HAMT approach:
https://en.wikipedia.org/wiki/Hash_array_mapped_trie

vs using a Vector, will be more performant, but maybe not... it will be fine.

In any case, if we make the details private, we can merge it with the right API and optimize later. I think we want a minimal API and as much as possible private.

Something like:
```
trait HashMap[K, +V] {
  def get(k: K)(implicit hash: Hash[K]): Option[V]
  def updated[V1 >: V](k: K, v: V1)(implicit hash: Hash[K]): HashMap[K, V1]
  // return the given value of the key and the state of the map without it
  def take(k: K)(implicit hash: Hash[K]): (Option[V], HashMap[K, V])
  def length: Long
  def iterator: Iterator[(K, V)]
 
  // without having to hash the key twice, transform a given key
  def transformKey[R, V1 >: V](k: K)(fn: Option[V1] => (R, Option[V1])): (R, HashMap[K, V1]) 

  // transform the values with the key given as input,but we cannot change the key
   // this can be faster since you don't change the structure or rehash the keys
  def mapWithKey[V1](fn: (K, V) => V1): HashMap[K, V1]
}

object HashMap {
  def empty[K, V]: HashMap[K, V]
}
```

sorenbug:
Ah! I couldn't find a `Hash` when looking for it earlier. Thank you!

sorenbug:
So, the new typeclass dependencies for some things seem a little odd. I haven't implemented a HAMT as of yet, but that'll be next. I'd like a review on my usage of `Hash` so far though.

sorenbug:
The Wikipedia page is _not_ helping for how to implement a HAMT, and my data structures course has nothing on it. What should I look at to implement one?

larsrh:
@sorenbug Thanks for following up! From my PoV, this implementation is ready for review (even without a HAMT implementation). Only thing missing now is tests for HashSet and HashMap. Ideally, the ones from AvlSet and AvlMap can be copied.

sorenbug:
Alright, I learned how to make a HAMT: It seems like I'd want to basically copy stdlib's `Vector` and index the tries by the bits of the `hash`. I'll see what I can do soon to add that here.

The current implementation has O(n) lookup because there's no resizing logic, and it's not a good implementation for an immutable HashMap anyways: usually this implementation is used for mutable HashMaps, even though adding an element is O(1).

johnynek:
I am closing this due to #534 but I appreciate you starting the ball rolling.

If you are still interested, I think there are still optimization opportunities in the hashmap code that was merged.

s5bug:
Yeah, as much as I wanted to finish this up I never really got around to understanding how to make better HashMaps. Thanks for keeping up with things!

PR #307, opened 2020-08-02
Scala 2.13.3

sullis:


larsrh:
Scala 2.13.3 appears to not like tut (#333)

larsrh:
Fixed by #355

PR #314, opened 2020-09-11
minor cleanup to remove deprecation on 2.13

johnynek:
I noticed Diet needlessly makes a stream to get an empty iterator.

We can directly make an empty iterator.


johnynek:
@larsrh what do you think of merging this minor change? It removes one deprecation warning on Stream on 2.13, and is probably marginally faster.

larsrh (approved):

PR #347, opened 2021-01-24
remove coverage from build

larsrh:


johnynek:
What was the motivation for this? Does this mean there are no coverage numbers any more?

larsrh:
I thought I saw a build error because of this, but I can't find it any more. I might be able to revert it.

Issue #346, opened 2021-01-23
Migrate to sbt-github-actions


larsrh:

brbrown25:
If you'd like I can assist in this one as well.

larsrh:
Sure thing.

larsrh:
#357

PR #355, opened 2021-02-03
switching to microsite from tut. #353

brbrown25:
@larsrh this should be all that's needed to get the site generated. I need to play around with it some more to get rid of it complaining about a non empty directory when running `makeMicrosite` as part of the `validate` command. But I figured I'd open this to make sure this is what you were looking for first.


larsrh:
Build is failing; looks like there is a mismatch between Scala versions in `.travis.yml` and `build.sbt`. Could you update the version in `.travis.yml`?

brbrown25:
> Build is failing; looks like there is a mismatch between Scala versions in `.travis.yml` and `build.sbt`. Could you update the version in `.travis.yml`?

That was my bad! I have updated the version.

brbrown25:
hmmm looks like it's still broken after that change I'll take a look again

brbrown25:
hmm https://github.com/scoverage/sbt-scoverage/issues/319#issuecomment-761228227 might be related

brbrown25:
@larsrh I've downgraded the scala version as it appears there is an issue with scoverage and binary compatability. with the version bump.

brbrown25:
Welp that broke resolve the micro site. I'll dig into this more over the weekend!

larsrh:
Thanks a ton for your effort. If it keeps failing, just ping me again and I can try having another look.

brbrown25:
@larsrh Thanks for the patience on this one. I forgot about some caveats of microsite when dealing with 2.12 and 2.13. I've updated the validate command to clean the docs folder so that you don't get a non empty directory error. I've also bumped the scala 2.12 and 2.13 versions as well as microsite so that this all resolves properly now. I went ahead and flattened this into a single commit so that the history can remain clean and you can accurately see what all was changed.  Let me know if you'd like any tweaks and thanks for the opportunity!

brbrown25:
Also of note, as it stands you may need to do some additional configuration outlined [here](https://47degrees.github.io/sbt-microsites/docs/publish-with-travis/ ) to get the publish working.

larsrh (approved):

larsrh:
Thanks a ton! Auto-publishing may be interesting after #346.

PR #414, opened 2021-08-02
Add Scalafmt

danicheg:
Rules are the same as in `cats`.


.scalafmt.conf
```diff
@@ -0,0 +1,12 @@
+version=2.7.5
```
osleonard, line 1: There is a new version we might as well go all the way to 3.0.0

danicheg: You are right, @osleonard. That work was done before 3.0.0 has been released. I'll update this PR soon.

danicheg:
I've updated Scalafmt to 3.0.1. It would be great if you can check this PR @gemelen, @lukoyanov, @osleonard.

osleonard (approved):
Looks like we have a flaky tests, but looks good running the tests again to see what could be causing the issue

lukoyanov:
LGTM.

lukoyanov (approved):

danicheg:
CI workflows were failed due to #425 that will be fixed in #428.

danicheg:
I will take another one try to green CI 🤕

core/src/main/scala/cats/collections/DisjointSets.scala
```diff
@@ -6,17 +6,18 @@ import cats.data.State.get
...
-    *
-    * @param a Set `a`
-    * @param b Set `b`
-    * @return (new [[DisjointSets]] with updated state,
-    *   `true` if Both labels are contained and joined
-    * )
-    */
+   * Joins two disjoint sets if both are contained by this [[DisjointSets]]
+   *
+   * @param a
```
gemelen, line 14: Probably worth bringing back (via inserting `-` between parts)

project/plugins.sbt
```diff
@@ -1,12 +1,14 @@
...
-addSbtPlugin("org.scala-js"       % "sbt-scalajs"                   % "1.7.0")
-addSbtPlugin("com.codecommit"     % "sbt-github-actions"            % "0.13.0")
+addSbtPlugin("com.47deg" % "sbt-microsites" % "1.3.4")
+addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.23")
+addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.3")
+addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
+addSbtPlugin("com.github.sbt" % "sbt-release" % "1.1.0")
+addSbtPlugin("io.crashbox" % "sbt-gpg" % "0.2.1")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.3")
+addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "1.0.0")
```
gemelen, line 8: This could be removed now, I guess.

osleonard (approved):

core/src/main/scala/cats/collections/DisjointSets.scala
danicheg: Scalafmt formats with a new line even with `-` 🤷🏻‍♂️

core/src/main/scala/cats/collections/TreeList.scala
```diff
@@ -658,12 +659,13 @@ object TreeList extends TreeListInstances0 {
           else {
             val h1 = h.updated(idx, a)
             // now rebuild the front of the list
-            Trees(front.reverse_:::(h1 :: tail))
+            Trees(front reverse_::: (h1 :: tail))
```
danicheg, line 662: To someone who interested in that fix - we've added `Scalafmt` with the `AvoidInfix` rewriting rule. So `Scalafmt` formatted `front reverse_::: (h1 :: tail)` as `front.reverse_:::(h1 :: tail)`. But that leads to behaviour changes for `reverse_:::` (see https://www.scala-lang.org/api/2.13.6/scala/collection/immutable/List.html#reverse_:::[B%3E:A](prefix:List[B]):List[B]).

gemelen (approved):

PR #416, opened 2021-08-03
Rm ScalaStyle

danicheg:
ScalaStyle probably is EOL. `cats` has already removed it (https://github.com/typelevel/cats/pull/3945). I propose to use only Scalafmt (#414) and reviewing to fix style things.


lukoyanov:
I reckon it makes sense to move forward with dropping `scalastyle` support for the aforementioned reasons, particularly as it was dropped in `cats` as well. 
What do you guys think @gemelen, @osleonard?

lukoyanov (approved):

osleonard (approved):

PR #428, opened 2021-09-03
Fix flakiness of `TreeListSuite`

### LINKED ISSUE #425: A flaky test `cats.collections.tests.TreeListSpec`
That test contains a flaky test suite "updated works" that fails from time to time. Recently I got an error:
```
 TestFailedException was thrown during property evaluation.
[info]     Message: Vector(1642340953, 0) did not equal Vector(0, 1642340953)
[info]     Location: (TreeListSpec.scala:244)
[info]     Occurred when passed generated values (
[info]       arg0 = TreeList(1378408384, 1642340953),
[info]       arg1 = 0, // 28 shrinks
[info]       arg2 = 0 // 63 shrinks
[info]     )
```

danicheg:
Closes #425

### DIFF
--- a/tests/src/test/scala/cats/collections/TreeListSuite.scala
+++ b/tests/src/test/scala/cats/collections/TreeListSuite.scala
@@ -244,7 +244,7 @@ class TreeListSuite extends DisciplineSuite {
   property("updated works")(forAll { (xs: TreeList[Int], v: Int, idx0: Long) =>
     def test(idx: Long): Unit = {
       val xs1 = xs.updatedOrThis(idx, v)
-      if (0 <= idx && idx < xs.size) {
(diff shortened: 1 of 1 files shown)


osleonard:
Nice fix, the issue is #414 wont be mergable, could we may be consolidate the change in #414 ?

danicheg:
Some observations: that test suite fails only in running a whole bunch of tests. When running it as a single suite via `testOnly` command - it's "green" all the time.

danicheg:
@osleonard probably, we can merge this first and then I'll rebase in #414 and rerun workflow to check that problem is solved. That test suite was flaky even earlier when tests used the `scalatest` as a test framework.

gemelen (approved):
Looks good!

osleonard (approved):

PR #466, opened 2021-12-04
Update algebra, algebra-laws to 2.7.0

gemelen:
Re-do
- https://github.com/typelevel/cats-collections/pull/459

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import ReleaseTransformations._
 val catsVersion = "2.7.0"
 val munitDisciplineVersion = "1.0.9"
 val scalacheckVersion = "1.15.4"
-val algebraVersion = "2.2.3"
+val algebraVersion = "2.7.0"
 val Scala212 = "2.12.15"
 val Scala213 = "2.13.7"
 val Scala3 = "3.0.2"
--- a/core/src/main/scala/cats/collections/Discrete.scala
+++ b/core/src/main/scala/cats/collections/Discrete.scala
@@ -3,7 +3,7 @@ package cats.collections
 /**
  * Represent discrete operations that can be performed on A
  */
-trait Discrete[@specialized(Specializable.Integral) A] {
+trait Discrete[@specialized(Specializable.Integral) A] extends Serializable {
 
   /**
    * Return the successor of x.


gemelen:
Ok, back to business :)
`DiesSuite` fails on serde tests of Diet object, specifically on all its 'Discrete` internal objects (ie not all components of internal state of `Diet` are serializable by default). If [`Discrete`](https://github.com/typelevel/cats-collections/blob/6b3daca35efbdd5a387a20b2271c5f730d5f4614/core/src/main/scala/cats/collections/Discrete.scala) trait is marked as a `Serializabe`, then tests pass successfully.

So, question is, are we allowed to make `Discrete` `Serializable` (I doubt so, but still want to ask)? 
@danicheg @osleonard  what do you think?

johnynek:
Why wouldn't you want to mark it serializable? Java Serialization is rather broken but useful for systems that serialize closures (like spark, flink, etc...)

I would add the `extends Serializable` without worrying very much.

danicheg:
If `MiMa` is not arguing against it, I vote to add that extends too.

gemelen:
> Why wouldn't you want to mark it serializable? Java Serialization is rather broken but useful for systems that serialize closures (like spark, flink, etc...)
> 
> I would add the `extends Serializable` without worrying very much.

FIrst of all because I consider `Discrete` as a public API (both in sense of being used by other parts in this library and by end-users), which I may not change that free.

On the other hand, if I'm too strict with this and we could make it `Serializable`, it makes intended upgrade easier.

armanbilge:
If it helps, basically all the typeclasses in Cats are serializable.

https://github.com/typelevel/cats/blob/490e3cfc3dbf6dc22156fa81140f51460ac4f7ac/kernel/src/main/scala/cats/kernel/Eq.scala#L18
https://github.com/typelevel/cats/blob/490e3cfc3dbf6dc22156fa81140f51460ac4f7ac/kernel/src/main/scala/cats/kernel/Semigroup.scala#L14
https://github.com/typelevel/cats/blob/490e3cfc3dbf6dc22156fa81140f51460ac4f7ac/core/src/main/scala/cats/Invariant.scala#L15

osleonard (approved):

Issue #480, opened 2022-01-17
Migrate to sbt-typelevel


lukoyanov:
According to the request from @armanbilge we need to migrate to sbt-typelevel.

Documentation is available here:
https://github.com/typelevel/sbt-typelevel/#readme

osleonard:
Will take of this

PR #483, opened 2022-01-22
Update sbt-github-actions to 0.14.2

gemelen:
Re-do of 
- #463

PR #485, opened 2022-01-23
Migrate to sbt-typelevel

armanbilge:
Closes https://github.com/typelevel/cats-collections/issues/480. Supersedes https://github.com/typelevel/cats-collections/pull/483.

Apologies to @osleonard if you were working on this already. This turned out be a very complex migration.
1. MiMa was not being run in CI, so there were some binary-incompatible changes in 0.9.x series. I'll add a comment about this below.
2. The benchmarks were not being compiled in CI, and currently do not compile in Scala 3, so I was unable to add them back to CI. This should be fixed in follow-up work and the benchmarks should be compiled in CI.
3. The scalac flags were updated to the standard used by most Typelevel projects. Some existing code could no longer compile and had to be modified.
4. Lots of warnings everywhere, so I disabled fatal warnings in CI. These should be fixed and re-enabled in follow-up work.
5. The diff seems big, but a lot of it is headers.


core/src/main/scala/cats/collections/Predicate.scala
```diff
@@ -76,7 +97,10 @@ trait PredicateInstances {
...
   }
 
-  implicit val predicateMonoidK: MonoidK[Predicate] = new MonoidK[Predicate] {
+  @deprecated("Retained for bincompat", "0.9.1")
+  def predicateInstance: MonoidK[Predicate] = predicateMonoidK
+
+  implicit def predicateMonoidK: MonoidK[Predicate] = new MonoidK[Predicate] {
     override def empty[A]: Predicate[A] = Predicate.empty
     override def combineK[A](l: Predicate[A], r: Predicate[A]): Predicate[A] = l.union(r)
   }
```
armanbilge, line 106: So we are in a bit of a hole wrt to bincompat. Bincompat with 0.9.0 was broken in https://github.com/typelevel/cats-collections/pull/269#discussion_r367173777 going into 0.9.1.

```
[error] core: Failed binary compatibility check against org.typelevel:cats-collections-core_2.12:0.9.0 (e:info.versionScheme=early-semver)! Found 5 potential problems
[error]  * static method predicateInstance()cats.MonoidK in class cats.collections.Predicate does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.collections.Predicate.predicateInstance")
[error]  * method predicateInstance()cats.MonoidK in object cats.collections.Predicate does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.collections.Predicate.predicateInstance")
[error]  * abstract method predicateInstance()cats.MonoidK in interface cats.collections.PredicateInstances does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.collections.PredicateInstances.predicateInstance")
[error]  * abstract synthetic method cats$collections$PredicateInstances$_setter_$predicateMonoidK_=(cats.MonoidK)Unit in interface cats.collections.PredicateInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.collections.PredicateInstances.cats$collections$PredicateInstances$_setter_$predicateMonoidK_=")
[error]  * abstract method predicateMonoidK()cats.MonoidK in interface cats.collections.PredicateInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.collections.PredicateInstances.predicateMonoidK")
```

I was able to restore it by making these changes here.

However, changing the `predicateMonoidK` from a `val` to a `def` now breaks bincompat on Scala 3:
```
[error] core: Failed binary compatibility check against org.typelevel:cats-collections-core_3:0.9.3 (e:info.versionScheme=early-semver)! Found 1 potential problems (filtered 34)
[error]  * synthetic static method $init$(cats.collections.PredicateInstances)Unit in interface cats.collections.PredicateInstances does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.collections.PredicateInstances.$init$")
```

Based on my reading of https://github.com/lightbend/mima/pull/448 this seems like a legitimate issue. However, it is relatively minor and Scala 3-only, so perhaps it would be okay to look the other way on this one ... I think fixing it would require splitting sources across Scala 2 / 3.

@johnynek would appreciate your thoughts here.

osleonard:
> Closes #480. Supersedes #483.
> 
> Apologies to @osleonard if you were working on this already. This turned out be a very complex migration.
> 
>     1. MiMa was not being run in CI, so there were some binary-incompatible changes in 0.9.x series. I'll add a comment about this below.
> 
>     2. The benchmarks were not being compiled in CI, and currently do not compile in Scala 3, so I was unable to add them back to CI. This should be fixed in follow-up work and the benchmarks should be compiled in CI.
> 
>     3. The scalac flags were updated to the standard used by most Typelevel projects. Some existing code could no longer compile and had to be modified.
> 
>     4. Lots of warnings everywhere, so I disabled fatal warnings in CI. These should be fixed and re-enabled in follow-up work.
> 
>     5. The diff seems big, but a lot of it is headers.

@armanbilge its ok started working on it yesterday, since you've made it happen it also fine

armanbilge:
@osleonard I'm sorry about that, I should have checked in first :( in any case, there's still much more work to be done to "fully" adopt sbt-typelevel in cats-collections, such as fixing all the warnings so you can enable fatal warnings in CI. I leave that to the maintainer team here :)

core/src/main/scala/cats/collections/Predicate.scala
johnynek: huh... I'd rather break bin-compat on scala 3 since I assume it is much less commonly used.

It is really weird that changing from val to def breaks binary compatibility... that's a real shame. Maybe that will be fixed in later scala 3?

armanbilge: I'm not sure how they can fix it binary-compatibly, unless they always generate dummy `$init$` methods. If `val`s generated an `$init$` method in the past, then they must continue to generate it going forward.

armanbilge:
Ok, this PR should be ready. I filtered the MiMa issue on Scala 3 and created these follow-ups:
- https://github.com/typelevel/cats-collections/issues/486
- https://github.com/typelevel/cats-collections/issues/487

gemelen (approved):
Looks good besides known issues

PR #533, opened 2022-07-27
Add cats.kernel.Hash-based CHAMP HashSet

DavidGregory084:
This is a port of typelevel/cats#4185 to cats-collections.


tests/src/test/scala/cats/collections/DefaultScalaCheckPropertyCheckConfig.scala
```diff
@@ -27,6 +27,7 @@ object DefaultScalaCheckPropertyCheckConfig {
   final val default: Parameters = Parameters.default
     .withMinSuccessfulTests(if (BuildInfo.isJvm) 50 else 5)
     .withMaxDiscardRatio(if (BuildInfo.isJvm) 5 else 50)
+    .withMaxSize(if (BuildInfo.isJvm) 10 else 5)
```
DavidGregory084, line 30: I added this to synchronise with [cats](https://github.com/typelevel/cats/blob/771c6c802f59c72dbc1be1898081c9c882ddfeb0/tests/shared/src/test/scala/cats/tests/CatsSuite.scala) - without it the tests take a really long time to run

gemelen:
As I understand, it's a latest version from Cats PR, so it's already went through a review regarding code and performance.
I have nothing to add to that review, as it's good to go from my point of view.

gemelen (dismissed):

armanbilge:
That's definitely the case for the `HashMap`, I think the `HashSet` (this PR) has had less direct review although I David did a great job applying review feedback from the `HashMap` also to the `HashSet`.

If the plans are for cats-collections to 1.0 in the near future (I hope so!) we should make sure everything is carefully reviewed :) I do think this one is in great shape!

johnynek (approved):

core/src/main/scala/cats/collections/HashSet.scala
```diff
@@ -0,0 +1,1144 @@
...
+      override def forall[A](fa: HashSet[A])(p: A => Boolean): Boolean = fa.iterator.forall(p)
+
+      override def count[A](fa: HashSet[A])(p: A => Boolean): Long =
+        fa.iterator.foldLeft(0L) { case (c, a) => if (p(a)) c + 1L else c }
+
+      def unorderedFoldMap[B, C](fa: HashSet[B])(f: B => C)(implicit C: CommutativeMonoid[C]): C =
+        C.combineAll(fa.iterator.map(f))
+    }
+
+  implicit def catsDataCommutativeMonoidForHashSet[A](implicit hash: Hash[A]): CommutativeMonoid[HashSet[A]] =
```
johnynek, line 1130: there is a question here as to which is the "correct" default monoid.

You can think of the set as an infinite vector with all entries that are present having the boolean value true, otherwise false (the contains function). You could OR those booleans (what this monoid is doing), but we could also AND them (intersection) or XOR them (exclusive or).

I tend to agree union is the most common, but we could also imagine having a non-implicit intersectionMonoid  which is also a `CommutativeMonoid`.

Lastly, you can implement a DistributiveLattice: https://github.com/typelevel/cats/blob/main/algebra-core/src/main/scala/algebra/lattice/DistributiveLattice.scala where join is union and meet is intersection.

core/src/main/scala-2.13+/cats/collections/compat/HashCompat.scala
```diff
@@ -0,0 +1,49 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.collections
+package compat
+
+import cats.kernel.Hash
+
+private[collections] class HashCompat {
```
DavidGregory084, line 27: Need to remember to apply copyright header here from scala/scala too

core/src/main/scala/cats/collections/HashSet.scala
```diff
@@ -0,0 +1,1144 @@
...
+
+    final override def next(): A = {
+      if (!hasNext) throw new NoSuchElementException
+      val value = currentNode.getValue(currentValuesIndex)
+      currentValuesIndex += 1
+      value
+    }
+  }
+
+  implicit val catsDataUnorderedFoldableForHashSet: UnorderedFoldable[HashSet] =
```
DavidGregory084, line 1111: Oh I need to rename these implicits too

core/src/main/scala/cats/collections/HashSet.scala
DavidGregory084: Yes that's worth exploring - I'm not sure now why I decided to implement `CommutativeMonoid` specifically here

DavidGregory084: I've added a `DistributiveLattice` instance in https://github.com/typelevel/cats-collections/pull/533/commits/aaa8a0a07ba2cbe98e436a50ea11f83a42d7853b. The default `CommutativeMonoid` instance is still the `HashSetUnionMonoid` but users can instantiate the `HashSetIntersectionMonoid` if they need it or use the `DistributiveLattice`.

gemelen (approved):

DavidGregory084:
Is there anything more to be done here? Anything I can do to move this forward?

PR #548, opened 2022-09-26
Add `BitSet.fromScalaRange`

armanbilge:
This will be helpful in idna4s /cc @isomarcte. The existing constructor forces to initialize a collection of integers.

### DIFF
--- a/core/src/main/scala/cats/collections/BitSet.scala
+++ b/core/src/main/scala/cats/collections/BitSet.scala
@@ -371,7 +371,7 @@ object BitSet {
   /**
    * Construct an immutable bitset from the given integer [[scala.collection.immutable.Range]].
    */
-  def apply(xs: scala.collection.immutable.Range): BitSet =
+  def fromScalaRange(xs: scala.collection.immutable.Range): BitSet =
     if (xs.isEmpty) Empty
     else {
       var bs = newEmpty(0)
--- a/core/src/main/scala/cats/collections/BitSet.scala
+++ b/core/src/main/scala/cats/collections/BitSet.scala
@@ -369,6 +369,19 @@ object BitSet {
     }
 
   /**
+   * Construct an immutable bitset from the given integer [[scala.collection.immutable.Range]].
+   */
+  def apply(xs: scala.collection.immutable.Range): BitSet =
+    if (xs.isEmpty) Empty
+    else {
+      var bs = newEmpty(0)
+      xs.foreach { i =>
+        bs = bs.mutableAdd(i)
+      }
+      bs
+    }
+
+  /**
    * Given a value (`n`), and offset (`o`) and a height (`h`), compute the array index used to store the given value's
    * bit.
    */
--- a/tests/src/test/scala/cats/collections/BitSetTest.scala
+++ b/tests/src/test/scala/cats/collections/BitSetTest.scala
@@ -375,11 +375,11 @@ object BitSetTest extends Properties("BitSet") {
     law(x, y, "--")(_ -- _)
   }
 
-  property("Range-constructor consistent with varargs-constructor") =
+  property("Bitset.fromScalaRange consistent with BitSet(...)") =
     forAll(Gen.chooseNum(0, Short.MaxValue.toInt), Gen.chooseNum(0, Short.MaxValue.toInt)) { (start, stop) =>
       forAll(Gen.chooseNum(1, (stop - start).abs.max(1))) { step =>
         val xs = start until stop by step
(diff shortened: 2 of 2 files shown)


core/src/main/scala/cats/collections/BitSet.scala
```diff
@@ -368,6 +368,19 @@ object BitSet {
       bs
     }
 
+  /**
+   * Construct an immutable bitset from the given integer [[scala.collection.immutable.Range]].
+   */
+  def apply(xs: scala.collection.immutable.Range): BitSet =
```
armanbilge, line 374: It's slightly awkward because cats-collections ships its own `Range[A]`. We could use that ... but the Scala `Range` is a bit more optimized and featureful for our needs here.

johnynek: if you want to, I think you could make this much more efficient.

Basically, there are five kinds of leafs here:

1. all zeros (can reuse a constant)
2. some number of zeros followed by all the rest ones.
3. some number of zeros followed by some runs, followed by all zeros
4. some number of ones followed by all zeros
5. all ones.

For very wide ranges, we could just create 1 leaf of type 2, 1 leaf of type 4, then in between we link to the same all one leaf.

This could be a huge win for ranges like `0 to (Int.MaxValue / 4)`

johnynek: the above is more complex when the step is not 1, but that's a very common case.

johnynek (approved):

core/src/main/scala/cats/collections/BitSet.scala
armanbilge: Ah yes, I suspected something like this might be possible. Thank you so much for writing that out, that is super helpful. Not sure I'll have time to work on it, but seems like a good followup.

Basically, for the idna4s use case the optimal constructor would be one that can take a list of (non-overlapping) ranges.

johnynek: I think union would be efficient enough to just do the above then union the list.

satorg: > but the Scala Range is a bit more optimized and featureful for our needs here.
What the features cats-collections' Range is missing comparing to the Scala's one?
Perhaps it would be better to allow constructing CC's Range from the Scala's and BitSet from the CC's one, wdyt?

One more thought.. Perhaps it makes sense to allow creation of `BitSet` from `Diet` (which is basically a set of ranges)? So it would be a more generic case available out there...

satorg: Also I'm not sure if it is a correct approach to overburden the `apply` ctor with `Range` type. Ranges don't look like a "natural" representation of BitSets to me. Perhaps constructor methods like `fromRange` would be a better bet for that.

armanbilge: > What the features cats-collections' Range is missing comparing to the Scala's one

I mean, nothing _critical_ ... basically Scala `Range` is specialized for `Int` and has an optimized `foreach` method. The Cats Collections `Range[A]` cannot do that optimization, by-definition. So we could accept that instead, and write a `while` loop ourselves to achieve the same effect. I was lazy /shrug

---

>One more thought.. Perhaps it makes sense to allow creation of BitSet from Diet (which is basically a set of ranges)? So it would be a more generic case available out there...

Are you sure? This is what the doc says.

> Diet is a binary search tree where each node contains a set of continuous values.

But I do like the idea.

---

> Also I'm not sure if it is a correct approach to overburden the `apply` ctor with `Range` type. Ranges don't look like a "natural" representation of BitSets to me. Perhaps constructor methods like `fromRange` would be a better bet for that.

Good point. I like overloads but they have fallen out of style for good reasons. Let me change this.

satorg: > Are you sure? This is what the doc says.
>> Diet is a binary search tree where each node contains a set of continuous values.

Hmm... Seems that the md docs are a bit misleading. A better wording is the `class Diet` scaladocs itself:
https://github.com/typelevel/cats-collections/blob/9ec73112b005d4711b313fef733f6649bb46198d/core/src/main/scala/cats/collections/Diet.scala#L34-L35

So yeah it appears to be a sort of "set of ranges" I think.

satorg (approved):

johnynek (approved):

gemelen (approved):

gemelen:
I guess I'll make a release after 2.13.10 would land.

PR #562, opened 2022-11-14
Fix compiler warnings in 'bench' module

satorg:
For #486

Issue #486, opened 2022-01-23
Fix the warnings and enable fatal warnings in CI


armanbilge:
Follow-up to https://github.com/typelevel/cats-collections/pull/485.

https://github.com/typelevel/cats-collections/blob/2bf3b430484c7788f3cadae5835690fb0cbd53c9/build.sbt#L17

satorg:
@armanbilge @gemelen may I take this one?

PATIENCE-ONAH:
Hi @armanbilge, it seems this issue has been resolved as there were no warnings after I ran test.  Do I go ahead and enable fatal warning or maybe I'm missing something.

armanbilge:
Hi @PATIENCE-ONAH, thanks for your interest!

> it seems this issue has been resolved

Unfortunately not, this issue has still not been resolved.

> there were no warnings after I ran test

The warnings occur before running tests, during the compilation phase. You can see them here in CI:
https://github.com/typelevel/cats-collections/actions/runs/18596764558/job/53024923908#step:12:25

satorg:
Ah yeah, this one... I started off briskly and fixed some of the files. Then got stuck with `TreeList`. I don't remember what the issue was exactly since it was quite long ago. Perhaps, there were differences between Scala2 and Scala3 compilers that I didn't manage to overcome.

PATIENCE-ONAH:
> Hi [@PATIENCE-ONAH](https://github.com/PATIENCE-ONAH), thanks for your interest!
> 
> > it seems this issue has been resolved
> 
> Unfortunately not, this issue has still not been resolved.
> 
> > there were no warnings after I ran test
> 
> The warnings occur before running tests, during the compilation phase. You can see them here in CI: https://github.com/typelevel/cats-collections/actions/runs/18596764558/job/53024923908#step:12:25

Okay, Thank you. I will look at it again

satorg:
Turns out it was quite straightforward: #768.
Still not sure what went wrong last time.

Note: there are still a handful of compiler warnings in the `tests` module. I can take on them another day.

cc @armanbilge @PATIENCE-ONAH

PR #569, opened 2022-12-13
Fix warnings in Diet

satorg:
For #486 (a follow-up to #562).

It was quite a tricky one. Most of the warnings in `Diet` were due to non-exhaustive matches in code like this:
https://github.com/typelevel/cats-collections/blob/4bdcca725d0be64d33a3d6ab1dfcaa958ee611b7/core/src/main/scala/cats/collections/Diet.scala#L50-L52

The root-cause was `EmptyDiet#unapply`:
https://github.com/typelevel/cats-collections/blob/4bdcca725d0be64d33a3d6ab1dfcaa958ee611b7/core/src/main/scala/cats/collections/Diet.scala#L319-L320

Apparently, the compiler could not figure out that `case EmptyDiet() =>` covers everything besides `DietNode`.
However there was a pitfall hidden: at first glance it might seem that it could be a solution to use the `EmptyDiet` object reference directly, i.e.:
```scala
this match {
  case EmptyDiet =>
  case DietNode(_, _, _) =>
```
However it does not work since `Diet` is invariant in its `[A]` whereas `EmptyDiet extends Diet[Nothing]`.
Therefore, the compiler rejects using descendants of `Diet[Nothing]` where `Diet[A]` is required.

A better option would be matching on the `EmptyDiet` object type directly, i.e.:
```scala
this match {
  case _: EmptyDiet.type =>
  case DietNode(_, _, _) =>
}
```
And it worked just fine for Scala2 (both 2.12 and 2.13), but stops working for Scala3 (because it is way better and more precise in regards to types, hehe). For more details see lampepfl/dotty#16503.

So the only approach I came up eventually with was introducing `final class EmptyDiet[A] extends Diet[A]` and making sure it has a singleton instance. Hence the most of the fixes in this PR.

### DIFF
--- a/bench/src/main/scala/cats/collections/bench/DietToIteratorRegressionBench.scala
+++ b/bench/src/main/scala/cats/collections/bench/DietToIteratorRegressionBench.scala
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2015 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.collections
+package bench
+
+import org.openjdk.jmh.annotations._
+import scala.util.Random
+
+@State(Scope.Benchmark)
+class DietToIteratorRegressionBench {
+  final private val randomSeed = 497169389666431579L
+
+  @Param(Array("100", "1000", "10000"))
+  var size: Int = _
+
+  @Param(Array("0", "8", "16", "24"))
+  var shiftBits: Int = _
+
+  var diet: Diet[Int] = Diet.empty[Int]
+
+  @Setup
+  def setup(): Unit = {
+    val random = new Random(randomSeed)
+
+    var i = 0
+    while (i < size) {
+      val s = random.nextInt()
+      val e = s + (random.nextInt() >> shiftBits)
+
+      diet += Range(s, e)
+      i += 1
+    }
+  }
+
+  @Benchmark
+  def toIterator(): Unit = {
+    diet.toIterator.foreach(Function.const(()))
+  }
+}
--- a/core/src/main/scala/cats/collections/Diet.scala
+++ b/core/src/main/scala/cats/collections/Diet.scala
@@ -48,7 +48,8 @@ sealed abstract class Diet[A] {
    * Returns true if x is a value is in the tree.
    */
   @tailrec final def contains(v: A)(implicit order: Order[A]): Boolean = this match {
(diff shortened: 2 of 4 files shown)


armanbilge:
Wow, nice!! I remember looking at this one as well and getting discouraged quite quickly 😅 sounds like you had a bit of a journey. Will take a closer look soon :)

bench/src/main/scala/cats/collections/bench/DietToIteratorRegressionBench.scala
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package cats.collections
+package bench
+
+import org.openjdk.jmh.annotations._
+import scala.util.Random
+
+@State(Scope.Benchmark)
+class DietToIteratorRegressionBench {
```
satorg, line 29: Since the implementation of `toIterator` has changed tremendously, I would like to make sure that there is no performance downgrade. Thus, with help of this new benchmark I measured throughput of the `toIterator` method for the old vs new versions. The numbers are below (Scala 2.13.10, openjdk64-11.0.16.1):

Old `toIterator`:
```
Benchmark                                 (shiftBits)  (size)   Mode  Cnt        Score        Error  Units
DietToIteratorRegressionBench.toIterator            0     100  thrpt   25  9454277.327 ± 202820.369  ops/s
DietToIteratorRegressionBench.toIterator            0    1000  thrpt   25  9180334.422 ± 168053.101  ops/s
DietToIteratorRegressionBench.toIterator            0   10000  thrpt   25  9435667.955 ± 175937.003  ops/s
DietToIteratorRegressionBench.toIterator            8     100  thrpt   25    13275.529 ±    116.664  ops/s
DietToIteratorRegressionBench.toIterator            8    1000  thrpt   25     2316.657 ±     21.123  ops/s
DietToIteratorRegressionBench.toIterator            8   10000  thrpt   25  9253691.385 ± 125554.986  ops/s
DietToIteratorRegressionBench.toIterator           16     100  thrpt   25    11878.898 ±     52.812  ops/s
DietToIteratorRegressionBench.toIterator           16    1000  thrpt   25      776.785 ±      4.678  ops/s
DietToIteratorRegressionBench.toIterator           16   10000  thrpt   25       62.205 ±      0.463  ops/s
DietToIteratorRegressionBench.toIterator           24     100  thrpt   25    11872.484 ±     13.432  ops/s
DietToIteratorRegressionBench.toIterator           24    1000  thrpt   25      773.976 ±      8.582  ops/s
DietToIteratorRegressionBench.toIterator           24   10000  thrpt   25       59.295 ±      0.426  ops/s
```
New `toIterator`:
```
Benchmark                                 (shiftBits)  (size)   Mode  Cnt         Score        Error  Units
DietToIteratorRegressionBench.toIterator            0     100  thrpt   25  29247017.785 ± 238590.018  ops/s
DietToIteratorRegressionBench.toIterator            0    1000  thrpt   25  28888641.861 ± 390222.792  ops/s
DietToIteratorRegressionBench.toIterator            0   10000  thrpt   25  29404256.921 ±  82854.894  ops/s
DietToIteratorRegressionBench.toIterator            8     100  thrpt   25    262861.156 ±   1205.228  ops/s
DietToIteratorRegressionBench.toIterator            8    1000  thrpt   25     58579.311 ±   1457.804  ops/s
DietToIteratorRegressionBench.toIterator            8   10000  thrpt   25  28011756.184 ± 528586.551  ops/s
DietToIteratorRegressionBench.toIterator           16     100  thrpt   25    234386.167 ±   4620.642  ops/s
DietToIteratorRegressionBench.toIterator           16    1000  thrpt   25     21461.643 ±    220.749  ops/s
DietToIteratorRegressionBench.toIterator           16   10000  thrpt   25      2107.130 ±     59.657  ops/s
DietToIteratorRegressionBench.toIterator           24     100  thrpt   25    235759.491 ±   2618.151  ops/s
DietToIteratorRegressionBench.toIterator           24    1000  thrpt   25     21929.967 ±    430.028  ops/s
DietToIteratorRegressionBench.toIterator           24   10000  thrpt   25      2121.011 ±     36.403  ops/s
```
So I would say that the performance looks quite a bit improved in there.

core/src/main/scala/cats/collections/Diet.scala
```diff
@@ -234,69 +236,67 @@ sealed abstract class Diet[A] {
...
   def toIterator: Iterator[Range[A]] = this match {
-    case EmptyDiet() => Iterator.empty
+    case _: EmptyDiet[_] => Iterator.empty
+
     case DietNode(focus, left, right) =>
-      left.toIterator.toStream
-        .append(Seq(focus))
-        .append(right.toIterator)
-        .iterator
+      left.toIterator ++ Iterator.single(focus) ++ right.toIterator
```
satorg, line 298: Actually, I didn't want to optimize it (at least, not in this PR), but I just had to get rid of warnings about `Stream` deprecations. Turned out this is the simplest way I figured out so far.

satorg:
> sounds like you had a bit of a journey.

Yeah.. I learned quite a lot about how Scala3 compiler works. Especially in regards to GADTs :)

gemelen (approved):

PR #570, opened 2022-12-14
fix more bench warnings

satorg:
For #486
Follow-up to #562
Somehow I overlooked some of the warnings in the **bench** module.

### DIFF
--- a/bench/src/main/scala-2/cats/collections/bench/BitSetBenchmarks.scala
+++ b/bench/src/main/scala-2/cats/collections/bench/BitSetBenchmarks.scala
@@ -44,13 +44,13 @@ class BitSetBenchmarks {
   var size: Int = _
   var indices: Array[Int] = _
 
-  var values: Array[Int] = _
+  var values: Vector[Int] = _
   var sci: immutable.BitSet = _
   var scm: mutable.BitSet = _
   var iset: immutable.Set[Int] = _
   var ccbs: _root_.cats.collections.BitSet = _
 
-  var values2: Array[Int] = _
+  var values2: Vector[Int] = _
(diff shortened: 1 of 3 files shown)


bench/src/main/scala-2/cats/collections/bench/BitSetBenchmarks.scala
```diff
@@ -44,13 +44,13 @@ class BitSetBenchmarks {
   var size: Int = _
   var indices: Array[Int] = _
 
-  var values: Array[Int] = _
+  var values: Vector[Int] = _
```
satorg, line 47: Starting with v2.13 the compiler complains, if `Array` is passed as a vararg parameter (like `immutable.BitSet(values: _*)` ) because it involves an implicit conversion to `Seq` and hence the entire array gets copied for that.

Changing to `Vector` because it is already `Seq` so no conversion is necessary.

bench/src/main/scala-2/cats/collections/bench/DietAddBench.scala
```diff
@@ -77,16 +77,16 @@ class DietAddBench extends BestCaseRangesList {
 
   @Benchmark
   def scalazDievAdd(): Unit = {
-    scalazValues.foldLeft(Diev.empty[Int])((d, r) => d + r)
+    scalazValues.foldLeft(Diev.empty[Int])((d, r) => d + r): Unit
```
satorg, line 80: Scala complains here and below that a non-unit value gets discarded.
Ascribing the expression type to `Unit` helps to calm it down.

gemelen (approved):

Issue #277, opened 2020-02-02
1.0


larsrh:
What is missing for a 1.0 release?

@johnynek argued that we should give bincompat guarantees, so I figure it's time for a 1.0 series.

armanbilge:
Before 1.0, we should switch the algebra dependencies to the forthcoming cats-algebra.

armanbilge:
I'd be happy to help 1.0 to materialize. In preparation for this, I think at least a couple of the PRs are 90% baked so can be finished up, and certainly a few of the open issues can be addressed. How long is the goal to continue publishing for Scala 2.12, can it be dropped for 1.0?

johnynek:
I chimed in on the cats-algebra thread. I'm very -1 on renaming the packages. And you can see a good example here.

Now we are saying, let's also break compatibility in cats-collections just so we can change the package name. I'm -1.

I think we should keep compatibility in algebra as we move the code into cats, and we should publish 1.0 here as is without blocking on that release since it will be compatible (especially for all the trait definitions).

gemelen:
For ease of reference: https://github.com/typelevel/cats/pull/3918

armanbilge:
@gemelen FYI We moved algebra into cats binary-compatibly (this was the right thing to do) so there is no need for any changes here.

gemelen:
Yep, I found out that it's a part of the cats now, but thanks for the info anyway.

armanbilge:
@gemelen I think it would be fantastic to start working seriously towards 1.0. I think two very good first steps toward this would be:
- https://github.com/typelevel/cats-collections/issues/486
- https://github.com/typelevel/cats-collections/issues/456

Then we can start triaging for 1.0. I'm happy to help review PRs where I can, please @ me.

armanbilge:
Nice job releasing v0.9.4 🎉

Next step towards 1.0: does anyone have time to work on https://github.com/typelevel/cats-collections/issues/486? :) It's not strictly necessary per-se, but IMHO fatal warnings enabled shows a commitment to code quality.

armanbilge:
Another update on this: I am becoming increasingly convinced of the importance of releasing a Cats Collections 1.0 to support several projects across the Typelevel ecosystem.

[...]

PR #571, opened 2022-12-21
keep fixing warnings in 'core'

satorg:
For #486 (a follow-up to #569).


core/src/main/scala/cats/collections/Heap.scala
```diff
@@ -308,11 +308,7 @@ object Heap {
     override def unbalanced: Boolean = size < (1L << height) - 1L
   }
 
-  private[collections] case object Leaf extends Heap[Nothing] {
-    def apply[A](): Heap[A] = this.asInstanceOf[Heap[A]]
-
-    def unapply[A](heap: Heap[A]): Boolean = heap.isEmpty
-
+  final private[collections] case class Leaf[A] private () extends Heap[A] {
```
satorg, line 311: Pretty much the same fix as it was in `Diet.scala` for the "match may not be exhaustive" warnings.

satorg:
Interesting.. There are no tests reported as failed but rather
> FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

https://github.com/typelevel/cats-collections/actions/runs/3751710955/jobs/6373035392#step:14:672

armanbilge:
Right, that means the Node.js VM crashed. Maybe one of the generators is making too much data?

satorg:
> Maybe one of the generators is making too much data?

Actually there was a mistake I made in the `Range.toIterator` implementation when was getting rid of `Stream`. The issue led to infinite loop in certain (quite rare) circumstances. A funny thing about it is that it was happening not in `RangeSuite` but rather in `DietSuite`.

However, when I started debugging it I noticed that there are similar issues with `Range.foreach` along with `foldLeft` and `foldRight`. The latter bug slipped through unnoticed because there were no tests that could reproduce necessary conditions for the bug to show up. So I decided to rewrite `RangeSuite` to make it using **scalacheck** and cover as many possible case as I could come up with.

core/src/main/scala/cats/collections/Range.scala
```diff
@@ -103,14 +104,18 @@ final case class Range[A](val start: A, val end: A) {
...
     var i = start
-    while (order.lteqv(i, end)) {
+    // Visit values in range [start, end), i.e. excluding `end`.
+    while (order.lt(i, end)) {
       f(i)
       i = discrete.succ(i)
     }
+    // Visit the last (or the only) value, if any.
+    if (order.eqv(i, end))
+      f(i)
```
satorg, line 118: The previous implementation was resulting to infinite loops in cases like
```
Range(123.toByte, Byte.MaxValue).foreach(_ => ())
```
I.e. due to integer overflows that could not be caught previously.

satorg:
Hmm...
```
[error]  * the type hierarchy of object cats.collections.PairingHeap#Leaf is different in current version. Missing types {cats.collections.PairingHeap}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("cats.collections.PairingHeap$Leaf$")
```
not sure, why Mima gets triggered on `PairingHeap.Leaf` which is private and always has been that way...

Moreover, I did pretty much the same trick for `Diet` last time and there were no Mima errors about it.
@armanbilge do you have a clue by chance?

armanbilge:
> not sure, why Mima gets triggered on `PairingHeap.Leaf` which is private and always has been that way...

MiMa has some issues sometimes with private classes inside objects. It should be safe to exclude.

armanbilge:
Hmm, I see MiMa is complaining about many things. We've already discussed breaking bincompat in https://github.com/typelevel/cats-collections/issues/277#issuecomment-1315664123, maybe we should just bump the base version to 1.0?

satorg:
@armanbilge 
> Hmm, I see MiMa is complaining about many things.

They are all about `PairingHeap.Leaf` and its internals. So seems I managed to calm Mima down with just two filters.

But yeah, it is quite weird that we have to have those filters for it after all. The only difference between `Diet.EmptyNode` and `PairingHeap.Leaf` I can spot is that the former's visibility is `private[collections]` wheres the latter's is just `private`. I can guess that the compiler may optimize private members out more aggressively comparing to package-local ones. But it's just my guess – not sure if I want to dig into it further :)

armanbilge:
> former's visibility is `private[collections]` wheres the latter's is just `private`

Yes, I've noticed similar behavior too. My guess is that `private` is Java-style package private (i.e. non-public) but `private[package]` is Scala-private but Java-public. Which is why MiMa considers them differently.

gemelen (approved):

PR #583, opened 2023-02-23
Add overlaps function for Range collection

keuhdall:
Hi !
I just came across this project and was planning to use the Range collection to check if a Range of Date was overlapping another one, bit it seems that such a function does not exist at the moment, so I tried implementing one myself, let me know if there's anything I should change :slightly_smiling_face:

### DIFF
--- a/core/src/main/scala/cats/collections/Range.scala
+++ b/core/src/main/scala/cats/collections/Range.scala
@@ -74,6 +74,12 @@ final case class Range[A](start: A, end: A) {
     order.lteqv(start, range.start) && order.gteqv(end, range.end)
 
   /**
+   * Verify whether or not the passed range overlaps the current range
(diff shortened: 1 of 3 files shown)


satorg:
Thank you for the PR!
Would it be possible (and make sense) to add a test that would check for consistency between `overlaps` and `&` methods please? I.e. if `overlaps` results to `true` then `&` should return `Some` for the same couple of ranges. Otherwise, it should return `None`.

keuhdall:
I just added a test for that, let me know if that's good enough, I'm pretty new at property-based testing, so I'm not sure if my way of doing things is very good :sweat_smile:

gemelen:
@keuhdall Thank you for your contribution

gemelen (approved):

PR #699, opened 2024-09-25
Core deps upgrade

gemelen:


gemelen:
Incorporates #691

gemelen:
@armanbilge Hi! Do you mind to advice, if it is viable to silence MiMA here, as we're jumping from 0.4 to 0.5 in native path and there is no previous artifacts, so the binary check fails on missing artifacts. Then we'd cut this with the 0.9.9 release and bring MiMA check back?

armanbilge:
@gemelen thanks for checking, the correct way to handle this is to bump the version here.

https://github.com/typelevel/cats-collections/blob/575824da37bc884b6f2125604ec0c86cd043a236/build.sbt#L46

armanbilge (approved):
thanks!

PR #794, opened 2026-02-09
Scala 3.8+ accommodations

gemelen:
Upgrade to 3.8.1 requires JVM 17+, which should be handled separately in a build process.
Also, it seems worth it to introduce a Scala 3 LTS version into the pipeline.


gemelen:
@durban totally valid point about publish, but before that stage it's useful to see the difference in the build process - at least at the moment of change to a newer scalac. That's why it's a draft and likely to be this way, as a testing ground.

PR #801, opened 2026-02-27
Add ArbitraryDequeue to scalacheck module and wire ArbitraryTreeList

krrishverma1805-web:
## What this PR does

- Adds a new `ArbitraryDequeue` instance to the `cats-collections-scalacheck` module, making `Arbitrary[Dequeue[A]]` available to downstream users.
- Wires the existing `ArbitraryTreeList` into `AllArbitrary` and the package object, which was previously defined but not included.

## Why

The `TODO.md` notes: *"move arbitrary instances somewhere that one can import into other projects"*.

Currently, `Arbitrary[Dequeue[A]]` is only defined inline in `DequeueSuite` and is not available to downstream library users who may want to write property-based tests involving `Dequeue`. This PR makes it available through the standard import paths:

```scala
import cats.collections.arbitrary.dequeue._
// or
import cats.collections.arbitrary.all._


krrishverma1805-web:
I've completed the onboarding contribution: 
https://github.com/typelevel/gsoc-onboarding/pull/11

Issue #821, opened 2026-05-15
Why is UnorderedTraverseForHashMap requiring a Hash constraint


loladenney:
The hash constraint is expanded to an implicit evidence parameter with type Hash[K], which is never used. Every method defined in the body of this implicit conversion definition has a parameter fa of type HashMap[K, A], from which the constraint K:Hash can be obtained. Furthermore for binary compatibility reasons any additions methods defined must start from existing ones so any future method will also have the same property. The implicit unused evidence parameter seems like it will never see any use, so it might be best to remove it (from what I understand this mean to mark the current one deprecated and have it point to the new one without the constraint).

https://github.com/typelevel/cats-collections/blob/943883f35b2506c0e3a33c9911481a298d9171c3/core/src/main/scala/cats/collections/HashMap.scala#L1429-L1431

PR #822, opened 2026-05-15
Suppress warnings

loladenney:
Warning in HashMap for catsCollectionsUnorderedTraverseForHashMap, it was forcing the compiler to generate an evidence parameter of type Hash[K], but this evidence isn't needed or used, causing an unused parameter warning. The fix was to manually de-sugar and mark the evidence parameter with \@unused

This is only a suppression for the issue described in #821, which would address the warning directly.

I plan to silence more warnings in this pr, this is only for a single warning.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,6 @@ ThisBuild / startYear := Some(2015)
 
 ThisBuild / crossScalaVersions := Seq(Scala3, Scala213, Scala212)
 ThisBuild / tlVersionIntroduced := Map("3" -> "0.9.3")
-ThisBuild / tlFatalWarnings := false
 ThisBuild / tlCiReleaseBranches := Seq("master")
 ThisBuild / tlSitePublishBranch := Some("master")
 ThisBuild / githubWorkflowJavaVersions := Seq("8", "21").map(JavaSpec.temurin)
--- a/tests/src/test/scala/cats/collections/DequeueSuite.scala
+++ b/tests/src/test/scala/cats/collections/DequeueSuite.scala
@@ -131,7 +131,7 @@ class DequeueSuite extends DisciplineSuite {
   })
 
   property("toList/toStream consistency")(forAll { (q: Dequeue[Int]) =>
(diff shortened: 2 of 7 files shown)


armanbilge (dismissed):

tests/src/test/scala/cats/collections/DequeueSuite.scala
```diff
@@ -131,7 +131,7 @@ class DequeueSuite extends DisciplineSuite {
   })
 
   property("toList/toStream consistency")(forAll { (q: Dequeue[Int]) =>
```
armanbilge:
```suggestion
  property("toList/toVector consistency")(forAll { (q: Dequeue[Int]) =>
```

build.sbt
```diff
@@ -15,7 +15,6 @@ ThisBuild / startYear := Some(2015)
 
 ThisBuild / crossScalaVersions := Seq(Scala3, Scala213, Scala212)
 ThisBuild / tlVersionIntroduced := Map("3" -> "0.9.3")
-ThisBuild / tlFatalWarnings := false
```
armanbilge, line 18: 🎉

tests/src/test/scala/cats/collections/DequeueSuite.scala
```diff
@@ -130,8 +130,8 @@ class DequeueSuite extends DisciplineSuite {
     assertEquals(q.toList, Foldable[Dequeue].toList(q))
   })
 
-  property("toList/toStream consistency")(forAll { (q: Dequeue[Int]) =>
-    assertEquals(q.toList, q.to[Stream, Int].toList)
+  property("toList/toVector consistency")(forAll { (q: Dequeue[Int]) =>
+    assertEquals(q.toList, q.to[Vector, Int].toList)
```
armanbilge, line 134: for posterity: this change is reasonable because what was being tested was not specific to `Stream` (or laziness). This is about testing the factory-based `to` method.

armanbilge (approved):
nice job, so great to have fatal warnings enabled. thanks!

Issue #634, opened 2023-09-12
a faster immutable stack/list


johnynek:
The canonical encoding of list in scala 2 would be:

```scala
sealed trait List[+A]
object List {
  case object Empty extends List[Nothing]
  case class NonEmpty[A](head: A, tail: List[A]) extends List[A]
}
```
but this implementation involves a lot of pointer chasing (to iterate N times into the List we need to dereference N times).

We know from work on immutable vectors that having blocks of arrays that we copy on change can give considerable improvements due to cache locality (and fewer derefs).

So one can imagine another list:

```scala
sealed trait List[+A] {
  def uncons: Option[(A, List[A])]
  def prepend(a: A): List[A]
}

object List {
  private val BlockSize = 16 // experiment on this
  private case object Empty extends List[Nothing] {
    def uncons = None
    def prepend(a: A) = {
      val ary = new Array[Any](BlockSize)
      val offset = BlockSize - 1
      ary(offset) = a
      Impl(offset, ary, Empty)
    }
  }
  private case class Impl[A](offset: Int, block: Array[Any], tail: List[A]) extends List[A] {
    def uncons = {
      val nextOffset = offset + 1
      val next = if (nextOffset == block.length) tail else Impl(nextOffset, block, tail)
      Some((block(offset).asInstanceOf[A], next)
   }
    def prepend(a: A) = {
      val ary = new Array[Any](BlockSize)
      if (offset > 0) {
        // copy the right side
        System.arraycopy(block, offset, ary, offset, BlockSize - offset) 
        val nextOffset = offset - 1
        ary(nextOffset) = a
        Impl(nextOffset, ary, tail)
      }
      else {
        val offset = BlockSize - 1
        ary(offset) = a
        Impl(offset, ary, this)
      }
    }
}
```
or something like this...

The idea is to copy the block on write.  Of course, `apply`, `map`, `foldLeft`, `foreach`, `iterator`, etc... could be significantly optimized since once you have a handle on the block iterating to the next value will be in cache.

With some appropriate benchmarks you could set the size of the block to some optimal value that I bet would be bigger than 1. It could be that this beats naive List in almost all cases, and when the List isn't too large it could compete with Vector (which isn't as fast as List for stack uses).

armanbilge:
I like this!

A couple other thoughts:

[...]

PR #825, opened 2026-05-28
Foundations for BList

loladenney:
building out the framework for the new datatype

### DIFF
--- a/core/src/main/scala/cats/collections/BList.scala
+++ b/core/src/main/scala/cats/collections/BList.scala
@@ -72,8 +72,8 @@ object BList {
       ary(offset) = a
       Impl(offset, ary, Empty)
     }
-    def headOption: Option[Nothing] = None
-    def tailOption: Option[Nothing] = None
+    def headOption: None.type = None
+    def tailOption: None.type = None
     def get(idx: Long): Option[Nothing] = None
     def getUnsafe(idx: Long): Nothing = throw new NoSuchElementException("invalid index")
     def lastOption: Option[Nothing] = None
--- a/core/src/main/scala/cats/collections/BList.scala
+++ b/core/src/main/scala/cats/collections/BList.scala
@@ -40,7 +40,7 @@ sealed abstract class BList[+A] {
   def concat[B >: A](l2: BList[B]): BList[B]
   def toList: List[A]
   def isEmpty: Boolean
-  def toIterator: Iterator[A]
+  def iterator: Iterator[A]
 
   final def ::[B >: A](a: B): BList[B] = prepend(a)
   final def ++[B >: A](l2: BList[B]): BList[B] = concat(l2)
@@ -93,7 +93,7 @@ object BList {
     def concat[B](l2: BList[B]): BList[B] = l2
     override def toList: List[Nothing] = Nil
     def isEmpty: Boolean = true
-    def toIterator: Iterator[Nothing] = Iterator.empty
+    def iterator: Iterator[Nothing] = Iterator.empty
     private[collections] def toStringInBlocks: String = "Empty"
 
     override def equals(other: Any): Boolean = other match {
@@ -101,7 +101,7 @@ object BList {
       case _             => false
     }
 
-    override def hashCode(): Int = Empty.toIterator.##
+    override def hashCode(): Int = Empty.iterator.##
 
   }
   sealed abstract class NonEmpty[+A] extends BList[A] {
@@ -136,19 +136,12 @@ object BList {
     override def equals(other: Any): Boolean =
       other match {
         case o: Impl[_] =>
-          val iterX = this.toIterator
-          val iterY = o.toIterator
-          while (iterX.hasNext && iterY.hasNext) {
-            if (iterX.next() != iterY.next()) { // not sure if i could be comparing with != here...
-              return false
-            }
-          }
-          iterX.hasNext == iterY.hasNext
+          this.iterator.sameElements(o.iterator.asInstanceOf[Iterator[A]])
         case _ => false
       }
 
     override def hashCode: Int =
-      MurmurHash3.orderedHash(this.toIterator)
+      MurmurHash3.orderedHash(this.iterator)
 
     def uncons: Some[(A, BList[A])] = {
       Some((block(offset), this.tail))
@@ -322,7 +315,7 @@ object BList {
     }
 
     def isEmpty: Boolean = false
-    def toIterator: Iterator[A] = new BListIterator(this)
+    def iterator: Iterator[A] = new BListIterator(this)
 
     private[collections] def toStringInBlocks: String = {
       val strb = new java.lang.StringBuilder
@@ -347,6 +340,28 @@ object BList {
       strb.append(")")
       strb.toString
     }
+    final private class BListIterator(var curNode: Impl[A @uncheckedVariance]) extends Iterator[A] {
+      var curOffset: Int = curNode.offset
+
+      def hasNext: Boolean = curOffset < BlockSize
+      def next(): A =
+        if (curOffset >= BlockSize) {
+          throw new NoSuchElementException
+        } else {
+          val next_elmt = curNode.block(curOffset)
+          curOffset += 1
+          if (curOffset >= BlockSize) {
+            // try to advance to next block
+            curNode.tailBList match {
+              case impl: Impl[A] =>
+                curOffset = impl.offset
+                curNode = impl
+              case Empty => // nothing, keep offset at blocksize
+            }
+          }
+          next_elmt
+        }
+    }
   }
 
   def fromList[A](l: List[A]): BList[A] =
@@ -365,28 +380,4 @@ object BList {
 
   def empty[A]: BList[A] = Empty
 
-  final class BListIterator[A](from: BList[A]) extends Iterator[A] {
-    private var curNode: BList[A] = from
-    private var curOffset: Int = if (!from.isEmpty) curNode.asInstanceOf[Impl[A]].offset else BlockSize
-
-    def hasNext: Boolean = !curNode.isEmpty
-    def next(): A =
-      curNode match {
-        case Empty         => throw new NoSuchElementException
-        case impl: Impl[A] => {
-          val next = impl.block(curOffset)
-          curOffset += 1
-          if (curOffset >= BlockSize) {
-            // advance to next block
-            curNode = impl.tailBList
-            curOffset = curNode match {
-              case Empty               => BlockSize
-              case impl_prime: Impl[A] => impl_prime.offset
-            }
-          }
-          next
-        }
-      }
-
-  }
 }
--- a/core/src/main/scala/cats/collections/BList.scala
+++ b/core/src/main/scala/cats/collections/BList.scala
@@ -430,19 +430,19 @@ object BList {
     }
 
     override def toList: List[A] = {
+      val builder = List.newBuilder[A]
       @tailrec
-      def loop(l: BList[A], acc: List[A]): List[A] =
+      def loop(l: BList[A]): List[A] =
         l match {
-          case Empty                    => acc
+          case Empty                    => builder.result()
           case impl: Impl[A] @unchecked =>
-            var newstuff: List[A] = List()
             // append valid things in the block to acc
             for (i <- impl.offset until BlockSize) {
-              newstuff = newstuff :+ impl.block(i)
+              builder += impl.block(i)
             }
-            loop(impl.tailBList, acc ++ newstuff)
+            loop(impl.tailBList)
         }
-      loop(this, List[A]())
+      loop(this)
     }
     def isEmpty: Boolean = false
 
--- a/core/src/main/scala/cats/collections/BList.scala
+++ b/core/src/main/scala/cats/collections/BList.scala
@@ -98,13 +98,15 @@ object BList {
     def tail: BList[A]
     def map[B](fn: A => B): BList.NonEmpty[B]
     def concat[B >: A](l2: BList[B]): BList.NonEmpty[B]
+    def uncons: Some[(A, BList[A])]
 
   }
 
   object NonEmpty {
     def apply[A](h: A, t: BList[A]): NonEmpty[A] =
       t.prepend(h) // maybe make this take arbitary number of args of same type and make a list out of them?
-    def unapply[A](l: BList[A]): Option[(A, BList[A])] = l.uncons
+    def unapply[A](l: NonEmpty[A]): Some[(A, BList[A])] =
+      l.uncons // can i put somethign to garuntee this still be Some
   }
 
   private object Impl {
@@ -116,7 +118,7 @@ object BList {
 
   // (maybe impl will be covariant or not)
   private case class Impl[A](offset: Int, block: Array[A], tailBList: BList[A]) extends NonEmpty[A] {
-    def uncons: Option[(A, BList[A])] = {
+    def uncons: Some[(A, BList[A])] = {
       val nextOffset = offset + 1
       val next = if (nextOffset == block.length) tailBList else Impl(nextOffset, block, tailBList)
       Some((block(offset), next))
@@ -190,7 +192,7 @@ object BList {
       @tailrec
       def go(self: Impl[A]): Some[A] = {
         self.tailBList match {
-          case Empty         => Some(block(BlockSize - 1))
+          case Empty         => Some(self.block(BlockSize - 1))
           case next: Impl[A] => go(next)
         }
       }
@@ -259,12 +261,12 @@ object BList {
       // TODO add special cases for if the block is a certain amount empty to shuffle things over
       // maybe use benchmarking to find out optimal number here??
 
-      // @tailrec   !!! not tailBListrec rn, could use cps but maybe ask first !!!
+      // @tailrec   !!! not tailBListrec rn, could use cps !!!
       def go(self: Impl[A]): BList.NonEmpty[B] = {
         self.tailBList match {
-          case Empty => Impl(offset, block.asInstanceOf[Array[B]], l2) // and the current block!!!
-          case next: Impl[A] => // maybe i shouldnt copy the block because no changes are occuring. any operation that needs to change blocks will copy, so realistically two lists can share an array if they both never mutate it
-            Impl(offset, block.asInstanceOf[Array[B]], go(next))
+          case Empty         => Impl(self.offset, self.block.asInstanceOf[Array[B]], l2)
+          case next: Impl[A] =>
+            Impl(self.offset, self.block.asInstanceOf[Array[B]], go(next))
         }
       }
 
@@ -336,6 +338,18 @@ object BList {
     }
      */
 
+  }
+  def fromList[A](l: List[A]): BList[A] =
+    fromListReverse(l.reverse)
+  def fromListReverse[A](l: List[A]): BList[A] = {
+    @tailrec
+    def go(l: List[A], acc: BList[A]): BList[A] = {
+      l match {
+        case Nil    => acc
+        case h :: t => go(t, acc.prepend(h))
+      }
+    }
+    go(l, empty)
   }
 
   def empty[A]: BList[A] = Empty
--- a/core/src/main/scala/cats/collections/BList.scala
+++ b/core/src/main/scala/cats/collections/BList.scala
@@ -33,15 +33,14 @@ sealed abstract class BList[+A] {
   def map[B](fn: A => B): BList[B]
   def foldLeft[B](init: B)(fn: (B, A) => B): B
   def drop(n: Long): BList[A]
-  def combineK [B >: A](l2: BList[B]): BList[B]
+  def combineK[B >: A](l2: BList[B]): BList[B]
   def toList: List[A]
-  //def strictFoldRight[B](fin: B)(fn: (A, B) => B): B
-  //final def take(n: Long): TreeList[A] =
+  // def strictFoldRight[B](fin: B)(fn: (A, B) => B): B
+  // final def take(n: Long): TreeList[A] =
 
   final def ::[B >: A](a: B): BList[B] = prepend(a)
 
   final def ++[B >: A](l2: BList[B]): BList[B] = combineK(l2)
-  
 
   override def toString: String = { // go back and optimize with blocks
     val strb = new java.lang.StringBuilder
@@ -66,7 +65,7 @@ object BList {
   private val BlockSize = 4 // test with different values
   case object Empty extends BList[Nothing] {
     def uncons = None
-    def prepend[A](a: A): BList[A] = { // why would we put the element at the end of a block? dont we want to fill from start out. in this case we would want head and tail pointer
+    def prepend[B >: Nothing](a: B): BList.NonEmpty[B] = { // why would we put the element at the end of a block? dont we want to fill from start out. in this case we would want head and tail pointer
       val ary = new Array[Any](BlockSize)
       val offset = BlockSize - 1
       ary(offset) = a
@@ -74,15 +73,15 @@ object BList {
     }
     def headOption: None.type = None
     def tailOption: None.type = None
-    def get(idx: Long): Option[Nothing] = None
+    def get(idx: Long): None.type = None
     def getUnsafe(idx: Long): Nothing = throw new NoSuchElementException("invalid index")
-    def lastOption: Option[Nothing] = None
+    def lastOption: None.type = None
     def size: Long = 0
     def map[B](fn: Nothing => B): BList[B] = Empty
     def foldLeft[B](acc: B)(fn: (B, Nothing) => B): B = acc
     def drop(n: Long): BList[Nothing] = Empty
     def combineK[B](l2: BList[B]): BList[B] = l2
-    def toList: List[Nothing] = Nil
+    override def toList: List[Nothing] = Nil
     // def strictFoldRight[B](acc: B)(fn: (Nothing, B) => B): B = acc
 
   }
@@ -108,13 +107,15 @@ object BList {
       val next = if (nextOffset == block.length) tail else Impl(nextOffset, block, tail)
       Some((block(offset).asInstanceOf[A], next))
     }
-    def prepend[B >: A](a: B): BList[B] = {
+    def prepend[B >: A](a: B): BList.NonEmpty[B] = {
 
       if (offset > 0) {
         // copy the right side
-        val ary = block.clone().asInstanceOf[Array[B]] // replaced with copyof method to prevent having to zero out the memory first
+        val ary = block
+          .clone()
+          .asInstanceOf[Array[B]] // replaced with copyof method to prevent having to zero out the memory first
         val nextOffset = offset - 1
-        ary(nextOffset) = a 
+        ary(nextOffset) = a
         Impl(nextOffset, ary, tail)
       } else {
         val ary = block.clone().asInstanceOf[Array[B]]
@@ -195,9 +196,9 @@ object BList {
       }
       loop(this, 0L)
     }
-    def map[B](fn: A => B): BList[B] = { 
+    def map[B](fn: A => B): BList[B] = {
       val ary = block.clone().asInstanceOf[Array[B]]
-      var i = 0
+      var i = offset
       while (i < ary.length) {
         ary(i) = fn(block(i))
         i += 1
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,236 @@
...
+        }
+        go(idx, this)
+      }
+    }
+    def getUnsafe(idx: Long): A = {
+      if (idx < 0) {
+        throw new IndexOutOfBoundsException
+      } else {
+        @tailrec
+        def go(idx: Long, l: BList[A]): A = {
```
satorg: Code-style nitpick: no need for `else` branch if `if` throws:

```suggestion
      if (idx < 0)
        throw new IndexOutOfBoundsException

      @tailrec
      def go(idx: Long, l: BList[A]): A = {
```
It'd reduce amount of braces and indentations in the code.

johnynek:
This is the same issue as #809  and #634 right?

A lot of the same discussion from #809 would be relevant here I think.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,298 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import scala.annotation.tailrec
+
+sealed abstract class BList[+A] {
+  def uncons: Option[(A, BList[A])]
+  def prepend[B >: A](a: B): BList[B]
```
satorg: I think we can say for sure that `prepend` operation always produces a non-empty collection:

```suggestion
  def prepend[B >: A](a: B): BList.NonEmpty[B]
```

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,298 @@
...
+  case object Empty extends BList[Nothing] {
+    def uncons = None
+    def prepend[A](a: A): BList[A] = { // why would we put the element at the end of a block? dont we want to fill from start out. in this case we would want head and tail pointer
+      val ary = new Array[Any](BlockSize)
+      val offset = BlockSize - 1
+      ary(offset) = a
+      Impl(offset, ary, Empty)
+    }
+    def headOption: Option[Nothing] = None
+    def tailOption: Option[Nothing] = None
```
satorg: For operations like these that always produce `None` for `Empty` instance we can try to make their return type simply `None.type`:

```suggestion
    def headOption: None.type = None
    def tailOption: None.type = None
```

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,298 @@
...
+      }
+      loop(this, 0L)
+    }
+    def map[B](fn: A => B): BList[B] = { 
+      val ary = block.clone().asInstanceOf[Array[B]]
+      var i = 0
+      while (i < ary.length) {
+        ary(i) = fn(block(i))
+        i += 1
+      }
```
satorg: Shouldn't we be starting from `offset` here (instead of 0)?

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,298 @@
...
+          case Empty =>
+            Empty
+          case Impl(offset, block, tail) =>
+            if (n >= BlockSize - offset) {
+              go(n - (BlockSize - offset), tail)
+            } else {
+              val m: Int = math.max(n.toInt, 0) // drop < 0 is the same as drop 0
+              val ary = block
+                .clone() // clone and new offset should do the trick because api doesnt expose the arrays so you can only access after offset
+              Impl(offset + m, ary, tail)
```
satorg: > clone and new offset should do the trick because api doesnt expose the arrays so you can only access after offset

However, if `BList` happens to contain references to some objects allocated on the heap, those references will be "locked" by this array and won't be GC-ed until the array is GC-ed itself, even if they are not referenced from anywhere else.

Therefore, if we reduce the size of the block after copying it, we have to clean up the unused space.

core/src/main/scala/cats/collections/BList.scala
loladenney: good catch, thank you!

core/src/main/scala/cats/collections/BList.scala
loladenney: Thank you for pointing this out. I have updated it accordingly

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,311 @@
...
+  private object Impl {
+    def apply[A](offset: Int, block: Array[A], tail: BList[A]): Impl[A] =
+      new Impl(offset, block, tail)
+    def apply[A](offset: Int, block: Array[Any], tail: BList[A]): Impl[A] =
+      new Impl(offset, block.asInstanceOf[Array[A]], tail)
+
+  }
+
+  // (maybe impl will be covariant or not)
+  private case class Impl[A](offset: Int, block: Array[A], tail: BList[A]) extends NonEmpty[A] {
```
johnynek: I think you can make this covariant if `block: Array[_ <: A]` that said, since we don't have a classtag for the type `A` and we may not want to require it, we may want to instead use `Array[Any]` and cast when we get out, which would require boxing primitives, but List also boxes primitives.

johnynek: `tail` here should probably be called `tailBList` or something to distinguish it from the logical `.tail` of the List.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,311 @@
...
+                go(idx - (BlockSize - offset), tail)
+              }
+          }
+        }
+        go(idx, this)
+      }
+    }
+    def getUnsafe(idx: Long): A = {
+      if (idx < 0)
+        throw new IndexOutOfBoundsException
```
johnynek, line 200: we are throwing a different kind of exception for `< 0` vs too large. Does `List` also do that? I would comment and follow List behavior here so it is fully compliant.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,311 @@
...
+      def loop(l: BList[A], acc: Long): Long = {
+        l match {
+          case Empty                 => acc
+          case Impl(offset, _, tail) => loop(tail, acc + (BlockSize - offset))
+        }
+      }
+      loop(this, 0L)
+    }
+    def map[B](fn: A => B): BList[B] = {
+      val ary = block.clone().asInstanceOf[Array[B]]
```
johnynek: I don't think we want this. We probably want to allocate a new `Array[Any]` that is the the same size as `ary`.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,311 @@
...
+
+    override def toList: List[A] = {
+      val builder = List.newBuilder[A]
+      @tailrec
+      def loop(l: BList[A]): List[A] =
+        l match {
+          case Empty                     => builder.result()
+          case Impl(offset, block, tail) =>
+            // append valid things in the block to acc
+            builder ++= block.slice(offset, BlockSize).toList
```
johnynek: I don't think we need the `toList` here and it will likely trigger a copy vs just a cheap iterator over the Array.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,311 @@
...
+        }
+      }
+      go(this, acc, fnblock)
+    }
+     */
+
+  }
+
+  def empty[A]: BList[A] = Empty
+  def unapply[A](l: BList[A]): Option[(A, BList[A])] = l.uncons
```
johnynek: I think this should be on the `object NonEmpty` inside `BList` and we should also have an `apply` method that builds a `NonEmpty` by prepending onto a BList. So `BList.NonEmpty(h, t)` can be used in construction and deconstruction.

satorg: Yes, makes sense. I think something like this should work:
```scala
object BList {
  object NonEmpty {
    def apply[A](h: A, t: BList[A]): NonEmpty[B] = ...
    def unapply[A](l: BList.NonEmpty[A]): Some[(A, BList[A])] = l.uncons
  }
}
```

core/src/main/scala/cats/collections/BList.scala
loladenney: It used to be Array[Any], but it ended up requiring so many uses of .asInstanceOf, and I didn't like how cluttered and confusing it made everything. Is there a good reason why we would want Impl to be covariant? If its worth it then I will try to find ways to contain the typecasts in a maintainable way.

core/src/main/scala/cats/collections/BList.scala
loladenney: Just checked and it seems to throw the same index out of bounds exception for index too small or too large!

core/src/main/scala/cats/collections/BList.scala
loladenney: I did it this way because clone doesn't zero out the space allocated to the array (while allocating a new array does), and since map will always write every valid (from offset until the end) cell in the array, the initial zeroing out of the allocated space will have no effect and none of the old data will remain. We can also assume that all blocks have zeroed out memory in any spots before offset, so there should be no issues with memory leaks from this. Considering this, should I still allocate a new array? There might be other benefits I haven't considered

core/src/main/scala/cats/collections/BList.scala
loladenney: What kinds of things do in the `object NonEmpty` vs the `object Impl`?

loladenney: Also, should the return type of unapply still keep the Option?

core/src/main/scala/cats/collections/BList.scala
satorg: I think Oscar makes a good point: `block.clone()` is only beneficial when we need to copy a substantial amount of data in a single shot. However, `map` never copies any data directly. Therefore, cloning the source array doesn’t make any sense here – we’ll overwrite every single cell in it anyway.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,315 @@
...
+    def combineK[B](l2: BList[B]): BList[B] = l2
+    override def toList: List[Nothing] = Nil
+    // def strictFoldRight[B](acc: B)(fn: (Nothing, B) => B): B = acc
+
+  }
+  sealed abstract class NonEmpty[+A] extends BList[A] {
+    // TODO can put methods in here that are only safe for nonempty (ex. head, reduce)
+    def head: A
+    def tail: BList[A]
+    // def reduce
```
satorg: I'd suggest establishing NonEmpty API here sooner than later. We don't need to repeat everything from `BList`, but only the methods that change their signatures for `NonEmpty`, e.g.:
```
def map[B](fn: A => B): BList.NonEmpty[B]
```
It would give us a clearer picture of all the requirements and constraints that we'll need for the implementation.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,315 @@
...
+  def headOption: Option[A]
+  def tailOption: Option[BList[A]]
+  def get(idx: Long): Option[A]
+  def getUnsafe(idx: Long): A
+  def lastOption: Option[A] // a bit more efficient than get(size-1)
+  def size: Long
+  def map[B](fn: A => B): BList[B]
+  def foldLeft[B](init: B)(fn: (B, A) => B): B
+  def drop(n: Long): BList[A]
+  def combineK[B >: A](l2: BList[B]): BList[B]
```
satorg: I'm not sure about this name – `combineK`. It is very generic and moreover `K` suffix suggests higher kinded types. The name comes from `SemigroupK`, which can be used for many different types, not just collections.

Sequence-like structures usually stick to names like `concat` or `++`. Alternatively, since `BList` tries to resemble the regular `List`, we can consider `:::` as well.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,315 @@
...
+
+import scala.annotation.tailrec
+
+sealed abstract class BList[+A] {
+  def uncons: Option[(A, BList[A])]
+  def prepend[B >: A](a: B): BList.NonEmpty[B]
+  def headOption: Option[A]
+  def tailOption: Option[BList[A]]
+  def get(idx: Long): Option[A]
+  def getUnsafe(idx: Long): A
```
satorg, line 34: `getUnsafe` or `unsafeGet`? I personally don't have any preference, but I've got a feeling that in the Typelevel ecosystem `unsafeOp` is used more often than `opUnsafe`. I might be wrong though.

core/src/main/scala/cats/collections/BList.scala
loladenney: I can rename it. I think I chose this one because Arman called it that when he gave me an initial list of methods to implement.

core/src/main/scala/cats/collections/BList.scala
loladenney: In cats-collections I'm generally seeing opUnsafe, but I'm also not vary familiar with the rest of the Typelevel ecosystem.

core/src/main/scala/cats/collections/BList.scala
satorg: > What kinds of things do in the object NonEmpty vs the object Impl?

`NonEmpty` is a part of the public API, whereas `Impl` hosts implementation details which we don't want to reveal.

> Also, should the return type of unapply still keep the Option?

That is a part of the trick: if we returned `Option[...]`, then Scala compiler might struggle on exhausiveness checks in some cases – because it doesn't know whether it resolves to `Some` or `None` at runtime. 

However, when we return `Some`, then we _reassure_ the compiler that the result does always exist:
```scala
val fooBList: BList[Int] = ...

fooBList match {
  case BList.NonEmpty(head, tail) => // only matches if our `fooBList` is a `NonEmpty` instance.
  case BList.Empty =>
  // or `case _ : BList.Empty.type =>` or just `case _ =>` – the `match` should be exhaustive in any case.
}
```
Minimal-ish example: https://scastie.scala-lang.org/smsfk232TM6C01PJonKThQ

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,351 @@
...
+        l.asInstanceOf[Impl[A]] match {
+          case Impl(_, block, tailBList) =>
+            tailBList match {
+              case Empty         => Some(block(BlockSize - 1))
+              case Impl(_, _, _) => go(tailBList)
+            }
+        }
+      }
+      go(this)
+    }
```
satorg, line 225: No need for `asInstanceOf` and other magic here. When `go` is called for the first time, we already have `Impl`, then we can just recurse to the same state, e.g.:
```scala
def lastOption: Some[A] = {
  @tailrec
  def go(self: Impl[A]): Some[A] = {
    self.tailBList match {
      case Empty         => Some(self.block(BlockSize - 1))
      case next: Impl[A] => go(next)
    }
  }
  go(this)
}
```

satorg: This rule is basically applicable to most of the other `go` functions inside `Impl` methods.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,356 @@
...
+
+    loop(true, this)
+    strb.append(")")
+    strb.toString
+  }
+
+}
+
+object BList {
+  private val BlockSize = 4 // test with different values
```
armanbilge:
```suggestion
  private[collections] final val BlockSize = 4 // test with different values
```

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,332 @@
...
+        ary(offset) = a
+        Impl(offset, ary, this)
+      }
+    }
+    def head: A = {
+      block(offset)
+    }
+    def tail: BList[A] = {
+      if (offset < BlockSize - 1) {
+        Impl(offset + 1, block, tailBList)
```
satorg: Are we sure we want to share arrays between two `BList` instances? It can results in discarded objects blocked from GC:
```scala
val largeThings = BList(largeThing1, largeThing2, largeThing3)
val res = largeThings.tail
```
In this case, `res` won't have `largeThing1` becasue it was discarded by `.tail`, but it will be still referenced by the same array that was shared between `largeThings` and `res`. Therefore, even if `largeThings` collection gets GC-ed, `largeThing1` will be kept in memory until `res` is GC-ed as well.

loladenney: the fix will be in the next commit. I also made the same fix for uncons.

core/src/main/scala/cats/collections/BList.scala
armanbilge: I think `getUnsafe` is more discoverable, since if someone starts typing `get` it will suggest the auto-complete.

More discoverable may be better or worse, depending on if we want people to use it or not haha.

> In cats-collections I'm generally seeing opUnsafe

Cool, I think it's reasonable to stick with that convention.

> but I've got a feeling that in the Typelevel ecosystem `unsafeOp` is used more often than `opUnsafe`.

I think that is the convention in Cats Effect. One notable difference there is that those operations are fundamentally `unsafe` and don't have safe version, unlike `get`.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,387 @@
...
+      new Impl(offset, block, tailBList)
+    def apply[A](offset: Int, block: Array[Any], tailBList: BList[A]): Impl[A] =
+      new Impl(offset, block.asInstanceOf[Array[A]], tailBList)
+    def unapply[A](l: Impl[A]): Some[(Int, Array[A], BList[A])] =
+      Some((l.offset, l.block, l.tailBList))
+  }
+
+  private class Impl[A](val offset: Int, val block: Array[A], val tailBList: BList[A]) extends NonEmpty[A] {
+
+    override def equals(other: Any): Boolean = {
```
satorg: While this implementation would probably work in general, it makes a lot of temporary heap allocations (for tuples and options). I’d suggest implementing it block‑by‑block rather than element‑by‑element, and avoiding allocations as much as possible. Something like this:
```scala
// Helper recursive class-level function.
@tailrec
private def isEqualTo(that: BList[A]): Boolean = {
  // We always know that `this` is `Impl` already.
  // Therefore we can safely short-circut to `false` if `that` is not.
  that match {
    case _: Empty.type => false
    case thatImpl: Impl[A] =>
        if (offset != thatImpl.offset || block.lengh != thatImpl.block.length)
          false
        else {
           // The offsets and block sizes are the same. Consequently,
           // we can use a boring yet efficient `while` loop over the array
           // with just a single
           var i = offset; // ..or `i = block.length - 1` if you will
           while { ... } // exits with `false` on the first mistmatch
           
           // If both blocks are equal, we can try to proceed with the next block:
           tailBList match {
              case _: Empty.type => thatImpl.isEmpty
              case nextImpl: Impl[A] =>
                nextImpl.isEqualTo(thatImpl.tailBList)
           }
        }
  }
}
```
Now, `override def equals(other: Any)` just needs to call our `isEqualTo` properly.

NOTES:
1. Instead of multi-level nesting, short-circuiting with `return false` would probably look better and cleaner.
2. `case _: Empty.type =>` can be more efficient compared to `case Empty =>`
3. Checks for `block.length` aren’t strictly necessary right now, but they will be if we ever decide to implement some kind of “adaptive” block sizes.
4. `do ... while` would fit even better our case, because every `Impl` is always non-empty. However, Scala 3 dropped `do ... while` in favor of `while { do side effects here then emit true or false }()` kind of thing.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,387 @@
...
+        case _ => false
+      }
+    }
+
+    override def hashCode(): Int = {
+      // uses only the valid elements in the first block
+      var res = 31
+      for (i <- this.offset until BlockSize) {
+        res = res * 31 + this.block(i).hashCode()
+      }
```
satorg: `while` would probably be more efficient that `for`:
```scala
var res = 31
var i = offset
while ( ... )
```
Besides, you already have `var res` here anyway.

core/src/main/scala/cats/collections/BList.scala
loladenney: Ok, so the issue I have with this implementation is that it checks for equality of blocks, which is stricter than the equality we want on ordered elements. Corresponding blocks of two BLists do not need to have the same offsets for the two lists to be equal. 
Consider the lists `BList(Block(1,2,3), Block(4,5), Empty)` and `BList(Block(1,2), Block(3,4,5), Empty)`. We want to say these are equal and I don't think isEqualTo will help with that, unless while checking equality we shift all the element to the left most spots as we go. That could be nice for reducing fragmentation, but can get expensive.

I could make isEqualTo compare arrays from a start index to an end index and do that where the blocks overlap...

loladenney: Update: I made a new equals that greedily compares within blocks. I think once I have some benchmarking up I might try to compare the performance between the two, but before then if anyone has thoughts on which might be faster I'd love to hear them.

scalacheck/src/main/scala/cats/collections/arbitrary/ArbitraryBList.scala
```diff
@@ -0,0 +1,60 @@
...
+package cats.collections.arbitrary
+
+import cats.collections.BList
+import org.scalacheck.Arbitrary
+import org.scalacheck.Arbitrary.arbitrary
+import org.scalacheck.Cogen
+import org.scalacheck.Gen
+
+trait ArbitraryBList {
+  def bListGen[A](implicit arb: Arbitrary[A], cogen: Cogen[A]): Gen[BList[A]] = {
```
satorg: Usually, `Gen` functions take `Gen` parameters, not `Arbitrary`. Moreover, `Cogen` param here looks a little strange – for a list-like structure it should not be necessary.

scalacheck/src/main/scala/cats/collections/arbitrary/ArbitraryBList.scala
```diff
@@ -0,0 +1,60 @@
...
+          } yield hs.prepend(a),
+          // concat
+          for {
+            xs <- Gen.resize(n / 4, bListGen[A])
+            ys <- Gen.resize(n / 4, bListGen[A])
+          } yield xs.concat(ys),
+          // map
+          Gen.resize(n / 2, bListGen[A]).flatMap { l =>
+            arbitrary[A => A].map(fn => l.map(fn))
+          }
```
satorg: This is the only place that requires `Cogen` instance, but I don't think we really need it:
```scala
// map     
Gen.resize(n / 2, bListGen[A]).map { l =>
  l.map(identity)
}
```
We can simply pass `identity` here, because it won't affect how `map` function works.

scalacheck/src/main/scala/cats/collections/arbitrary/ArbitraryBList.scala
loladenney: I based my implementation off the other arbitrary datatype implementations in the same folder, and it seems like none of their `Gen` functions have a `Gen` parameter, but they do have implicit `Arbitrary` parameters. 
As for the Cogen param, I will remove it.

scalacheck/src/main/scala/cats/collections/arbitrary/ArbitraryBList.scala
loladenney: Thank you! I made this change.

core/src/main/scala/cats/collections/BList.scala
satorg: > the issue I have with this implementation is that it checks for equality of blocks, which is stricter than the equality we want on ordered elements.

Good point. In that case I'd personally go with the `Iterator`-based approach.

I believe we do need to get `Iterator` here anyway, and having that, the `equals` and `hashCode` can be implemented in a very simple and straightforward way.

scalacheck/src/main/scala/cats/collections/arbitrary/ArbitraryBList.scala
satorg: Thanks for the clarification. It's somewhat surprising that `*Gen` functions are implemented that way in CatsCollections. The ScalaCheck itself follows a slightly different approach.

For example,
[stringOf](https://github.com/typelevel/scalacheck/blob/3b1e58f6f06cd540bd9acda7794c5fc51665866f/core/shared/src/main/scala/org/scalacheck/Gen.scala#L1194-L1196) takes `Gen` and returns `Gen`, whereas [arbString](https://github.com/typelevel/scalacheck/blob/3b1e58f6f06cd540bd9acda7794c5fc51665866f/core/shared/src/main/scala/org/scalacheck/Arbitrary.scala#L179-L181) is an implicit function that takes `implicit Arbitrary` (and uses `stringOf` under the hood).

Likewise,
[buildableOf](https://github.com/typelevel/scalacheck/blob/3b1e58f6f06cd540bd9acda7794c5fc51665866f/core/shared/src/main/scala/org/scalacheck/Gen.scala#L1007-L1010) takes `Gen[T]` explicitly and returns `Gen` as well, whereas [arbContainer](https://github.com/typelevel/scalacheck/blob/3b1e58f6f06cd540bd9acda7794c5fc51665866f/core/shared/src/main/scala/org/scalacheck/Arbitrary.scala#L387-L394) is an implicit that takes `Arbitrary[T]` implicitly and returns `Arbitrary`.

So I'd say – that was the original idea of ScalaCheck, whereas the current implementation in CatsCollections deviates from there quite a bit.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,392 @@
...
+  def getUnsafe(idx: Long): A
+  def lastOption: Option[A]
+  def size: Long
+  def map[B](fn: A => B): BList[B]
+  def foldLeft[B](init: B)(fn: (B, A) => B): B
+  def drop(n: Long): BList[A]
+  def concat[B >: A](l2: BList[B]): BList[B]
+  def toList: List[A]
+  def isEmpty: Boolean
+  def toIterator: Iterator[A]
```
satorg: Actually, `.toIterator` is deprecated in Scala Library in favor of `.iterator` starting with v2.13:
https://github.com/scala/scala/blob/b5e60a0853bbd53f9975050240199483d41bda92/src/library/scala/collection/IterableOnce.scala#L204-L205

So perhaps, `BList` should use name `iterator` here as well.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,392 @@
...
+    def getUnsafe(idx: Long): Nothing = throw new IndexOutOfBoundsException
+    def lastOption: None.type = None
+    def size: Long = 0
+    def map[B](fn: Nothing => B): BList[B] = Empty
+    def foldLeft[B](acc: B)(fn: (B, Nothing) => B): B = acc
+    def drop(n: Long): BList[Nothing] = Empty
+    def concat[B](l2: BList[B]): BList[B] = l2
+    override def toList: List[Nothing] = Nil
+    def isEmpty: Boolean = true
+    def toIterator: Iterator[Nothing] = new BListIterator(this)
```
satorg:
```suggestion
    def toIterator: Iterator[Nothing] = Iterator.empty
```

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,392 @@
...
+      other match {
+        case o: Impl[_] =>
+          val iterX = this.toIterator
+          val iterY = o.toIterator
+          while (iterX.hasNext && iterY.hasNext) {
+            if (iterX.next() != iterY.next()) { // not sure if i could be comparing with != here...
+              return false
+            }
+          }
+          iterX.hasNext == iterY.hasNext
```
satorg: Perhaps, it can simply call `iterX.sameElements(iterY)` (after casting to the same base element type):
https://github.com/scala/scala/blob/b5e60a0853bbd53f9975050240199483d41bda92/src/library/scala/collection/Iterator.scala#L841-L855

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,392 @@
...
+    }
+    go(l, empty)
+  }
+
+  def empty[A]: BList[A] = Empty
+
+  final class BListIterator[A](from: BList[A]) extends Iterator[A] {
+    private var curNode: BList[A] = from
+    private var curOffset: Int = if (!from.isEmpty) curNode.asInstanceOf[Impl[A]].offset else BlockSize
+
```
satorg: `BListIterator` can be `private`. In that case, no need to copy init params into private variables. Moreover, since `Empty.iterator` can simply return `Iterator.empty`, perhaps no need to pass common base `BList` here and simply work with `Impl` all along:
```scala
private final class BListIterator[A](
  var curNode: Impl[A],
  var curOffset: Int) extends Iterator[A] {

  def hasNext: Boolean =
    // Leave `curOffset` at `curNode.block.length` if `curNode` is the last one.
    curOffset < curNode.block.length

  def next: A = ???
```

loladenney: Thanks! Will do.

core/src/main/scala/cats/collections/BList.scala
loladenney: This seems cleaner, thanks!

core/src/main/scala/cats/collections/BList.scala
loladenney: made this change!

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,383 @@
...
+      ary(offset) = a
+      Impl(offset, ary, Empty)
+    }
+    def headOption: None.type = None
+    def tailOption: None.type = None
+    def get(idx: Long): None.type = None
+    def getUnsafe(idx: Long): Nothing = throw new IndexOutOfBoundsException
+    def lastOption: None.type = None
+    def size: Long = 0
+    def map[B](fn: Nothing => B): BList[B] = Empty
```
satorg, line 90: A few really minor nitpicks:
```suggestion
    def map[B](fn: Nothing => B): BList[B] = this
```

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,383 @@
...
+    def foldLeft[B](acc: B)(fn: (B, Nothing) => B): B = acc
+    def drop(n: Long): BList[Nothing] = Empty
```
satorg, line 92:
```suggestion
    def drop(n: Long): BList[Nothing] = this
```

armanbilge:
Note we are merging this PR into an intermediate branch `l/faster-immutable-list-634` to break up this large project into smaller stages. The project will culminate with a final PR merging everything into the main branch.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -0,0 +1,383 @@
...
+    def concat[B](l2: BList[B]): BList[B] = l2
+    override def toList: List[Nothing] = Nil
+    def isEmpty: Boolean = true
+    def iterator: Iterator[Nothing] = Iterator.empty
+    private[collections] def toStringInBlocks: String = "Empty"
+
+    override def equals(other: Any): Boolean = other match {
+      case _: Empty.type => true
+      case _             => false
+    }
```
satorg, line 102:
```suggestion
    override def equals(other: Any): Boolean = this eq other
```

loladenney: This change does not compile

loladenney: I changed it to `other.isInstanceOf[Empty.type]` but I don't know if that is better.

core/src/main/scala/cats/collections/BList.scala
loladenney: I changed everything to fit this pattern. it will be in the next pr about typeclasses

core/src/main/scala/cats/collections/BList.scala
satorg: Sure – it should work too, thanks.

PR #837, opened 2026-07-08
Typeclasses and other additions

loladenney:


core/src/main/scala/cats/collections/BList.scala
```diff
@@ -375,9 +729,188 @@ object BList {
...
+        }
+        fa match {
+          case Empty         => arg0.pure(BList.empty)
+          case impl: Impl[A] =>
+            arg0.map2(
+              traverseArray(impl.block.slice(impl.offset, BlockSize))(f).asInstanceOf[G[Array[B]]],
+              traverse(impl.tailBList)(f)
+            ) { (arrayB, tailBList) =>
+              Impl(impl.offset, new Array[Any](impl.offset) ++ arrayB, tailBList)
+            }
```
armanbilge: We can delegate to `nonEmptyTraverse`

loladenney: got it!

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -213,6 +362,37 @@ object BList {
...
+                                 (BlockSize - impl.offset) - idx
+                )
+                val lastBlockl1 = Impl(BlockSize - idx.toInt, ary1, Empty)
+                val l2 = Impl(impl.offset + idx.toInt, ary2, impl.tailBList)
+                Eval.now((lastBlockl1, l2))
+              }
+            } else { // split isn't in current block
+              Eval.defer(buildLists(idx - (BlockSize - impl.offset), impl.tailBList)).map {
+                _ match { case (l1_tail, l2) => (Impl(impl.offset, impl.block, l1_tail), l2) }
+              }
```
satorg: I'm not entirely sure what this `_ match { ...` is for. Perhaps you meant something like this:
```scala
Eval.defer(buildLists(idx - (BlockSize - impl.offset), impl.tailBList)).map {
  case (l1_tail, l2) => (Impl(impl.offset, impl.block, l1_tail), l2)
}
```

loladenney: Didn't realize the ` _ match` could be omitted. Thanks!

loladenney:
[BListresults.json](https://github.com/user-attachments/files/31275177/BListresults.json)
recent benchmarking results.

core/src/main/scala/cats/collections/BList.scala
```diff
@@ -227,52 +455,141 @@ object BList {
...
+          case Empty                 => Eval.now(Empty)
+          case impl: AbstractImpl[A] =>
+            val newblock = new Array[Any](BlockSize)
+            for (i <- impl.offset until BlockSize) {
+              newblock(i) = fn(impl.block(i))
+            }
+            Eval.defer(loop(impl.tailBList).map(Impl(impl.offset, newblock, _)))
+        }
+      }
+      loop(this).value.asInstanceOf[NonEmpty[B]]
```
armanbilge: I think we can eke out some more performance here, by building up the `BList` mutably from the front instead of using `Eval`. Same for `filter`.

In general, I think we may be able to replace uses of `Eval` with the builder.

loladenney: This change was going back to the Eval implementation to see if there were any improvements, but the builder implementation seems better so ill revert it.

